PR #17, opened 2019-11-14
fix typo

edson-junior:


edson-junior:
> @edson-junior Thanks Sherlock! 🔍

It was my pleasure! 😬

PR #22, opened 2019-12-30
feat(nohoist): add react-native packages to nohoist

renanmav:
Some packages shouldn’t be installed in the root `node_modules`, so I did a nohoist of these packages


renanmav:
Captura de Tela 2019-12-30 às 00 07 26

Some error with `react` module resolver on metro bundler. I’ll solve this later on.

renanmav:
I will close this is order to upgrade React Native to avoid hoisting

PR #23, opened 2020-03-31
Fix facebook login

belgamo:
When I tried to login with facebook, I received the error message `logInWithReadPermissions is not a function`. I think the version 0.10.0 of `react-native-fbsdk` uses `logInWithPermissions` function.

Reference: https://github.com/facebook/react-native-fbsdk/issues/566


wendelfreitas:
@belgamo you're right, thanks for this pr

PR #28, opened 2020-04-29
Add login page test

belgamo:


packages/mobile/__tests__/pages/Login.test.js
```diff
@@ -0,0 +1,83 @@
...
+    const { getByText, getByTestId } = render(component);
+
+    const btn = getByText('Entrar com facebook');
+
+    fireEvent.press(btn);
+
+    await waitForElement(() => getByTestId('loading'));
+  });
+
+  it('when login is successful', async () => {
```
kostadriano: When you gonna describe a scenario like "when whatever" you use `describe`. On `it` you put the expected response. Ex:

describe => when login is sucessful
it => Call expected function or show "blabla" message to user

packages/mobile/__tests__/pages/Login.test.js
```diff
@@ -0,0 +1,83 @@
...
+    fireEvent.press(btn);
+
+    await wait();
+
+    expect(handleLoginFacebook).toHaveBeenCalled();
+    expect(AsyncStorage.getItem).toHaveBeenCalled();
+    expect(AsyncStorage.setItem).toHaveBeenCalled();
+  });
+
+  it('when login is not successful', async () => {
```
kostadriano: Same

packages/mobile/__tests__/pages/Login.test.js
```diff
@@ -0,0 +1,83 @@
...
+describe('Login', () => {
+  const mockStore = configureStore();
+  const store = mockStore({});
+  const apolloStore = createApolloStore(USER_LOGIN_MUTATION);
+
+  const component = (
+    <Wrapper apolloStore={apolloStore} reduxStore={store}>
+      <Login navigation={navigation} />
+    </Wrapper>
+  );
```
nolleto: Maybe you extract this to a method like:

```js
const renderComponnent = () => {
  const mockStore = configureStore();
  const store = mockStore({});
  const apolloStore = createApolloStore(USER_LOGIN_MUTATION);
  const component = (
    <Wrapper apolloStore={apolloStore} reduxStore={store}>
      <Login nagivstion={navigation} />
    </Wrapper>
  );

  return render(component)
}

describe('Login', () => {
...
```

This way, you can call render component like this:
```js
it('...', () => {
  const { getByText } = renderComponnent();
  ...
```
and you will get a new component and store every time and avoid one test impact in another (in some cases).

packages/mobile/__tests__/pages/Login.test.js
belgamo: It makes sense! I'll improve that. Thanks!

packages/mobile/__tests__/pages/Login.test.js
belgamo: It looks good! Thanks!

packages/mobile/__tests__/pages/Login.test.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    it('handle properly the login flow', async () => {
+      const { getByText } = renderComponent();
+
+      const btn = getByText('Entrar com facebook');
+
+      fireEvent.press(btn);
+
+      await wait();
```
nolleto, line 60: What is the purpose of this empty `wait`? 🤔

packages/mobile/__tests__/pages/Login.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+      expect(handleLoginFacebook).toHaveBeenCalled();
+      expect(AsyncStorage.getItem).toHaveBeenCalled();
+      expect(AsyncStorage.setItem).toHaveBeenCalled();
```
nolleto, line 64: What do you think of one `expect` fo `it`?
This usually makes your test description better and if one of the test fails you will know what is not working.
```js
describe('when login is successful', () => {
  ...
  // you can put a better description if you want 😄 
  it('calls handleLoginFacebook method', async () => {
    const { getByText } = renderComponent();
    const btn = getByText('Entrar com facebook');

    fireEvent.press(btn);
    await wait();

    expect(handleLoginFacebook).toHaveBeenCalled();
  });

  it('calls AsyncStorage.getItem method', async () => {
    const { getByText } = renderComponent();
    const btn = getByText('Entrar com facebook');

    fireEvent.press(btn);
    await wait();

    expect(AsyncStorage.getItem).toHaveBeenCalled();
  });

  it('calls AsyncStorage.setItem method', async () => {
    const { getByText } = renderComponent();
    const btn = getByText('Entrar com facebook');

    fireEvent.press(btn);
    await wait();

    expect(AsyncStorage.setItem).toHaveBeenCalled();
  });
  ...
});
```
output **before** this refactor:
```
when login is successful 
  handle properly the login flow
```
output **after** this refactor:
```
when login is successful 
  calls handleLoginFacebook method
  calls AsyncStorage.getItem method
  calls AsyncStorage.setItem method
```

Now, if `expect(AsyncStorage.getItem).toHaveBeenCalled();` fails you will know what exactly is breaking in your test. 
Before this refactor 
```
expect(handleLoginFacebook).toHaveBeenCalled(); ✅ 
expect(AsyncStorage.getItem).toHaveBeenCalled(); 🚫 
expect(AsyncStorage.setItem).toHaveBeenCalled(); ✅ 
```
you will have to debug the test to know which expect fails.

So, what do you think? 😄

belgamo: First of all, thank you for your attention to my PR.

I'am not very experienced in testing and I understand your point of view, but I don't totally agree. In this case, I think it's better put all the assertions together, because the following snippet sets up them:
```
const { getByText } = renderComponent();
const btn = getByText('Entrar com facebook');

fireEvent.press(btn);
```

Besides that, this:
```
when login is successful 
  handle properly the login flow
```
wouldn't be a problem because jest shows exactly what expectation has failed.

Take a look at [here](https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test#comment679567_7829) and [here](https://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test#comment368610_7823).

So, what do you think?

packages/mobile/__tests__/pages/Login.test.js
belgamo: The purpose of this is await the mock promises has been resolved.

Reference: https://www.native-testing-library.com/docs/api-async

packages/mobile/__tests__/pages/Login.test.js
nolleto: Hi, I had forgotten that jest you tell you which `expect` fails :palm_face:

Just commenting more, about this
```js
const { getByText } = renderComponent();
const btn = getByText('Entrar com facebook');

fireEvent.press(btn);
```
IMHO, I don't think it's a problem since this is o unit test and this setup shouldn't take more than 1 sec.
I usually like to keep one `expect` for `it` to let the test more described, almost like a doc for the component behavior. In my experience with tests I saw some big `it`'s with a lot of `expect` and poorly descriptions like:
```js
describe('Login Component', () => {
  it('renders fields', () => {
    const { getByLabelText, queryByText } = renderComponentLogic()

    expect(getByLabelText('Username').value).toBe('')
    expect(getByLabelText('Password').value).toBe('')
    expect(queryByText('Sign In')).toBeTruthy()
  })
})
```
IHMO, I prefer something like
```js
describe('Login Component', () => {
  it('renders username field empty', () => {
    const { getByLabelText } = renderComponentLogic()

    expect(getByLabelText('Username').value).toBe('')
  })

  it('renders password field empty', () => {
    const { getByLabelText } = renderComponentLogic()

    expect(getByLabelText('Password').value).toBe('')
  })

  it('renders sign in button', () => {
    const { queryByText } = renderComponentLogic()

    expect(queryByText('Sign In')).toBeTruthy()
  })
})
```
This way in this example, I know exactly what is the behave of this component and what it will render reading the terminal messages or reading the code.

But, you can keep your code because I don't know the pattern of the tests for this repo. Maybe would be better the Author of this repo check this tests 😄

PR #46, opened 2020-05-31
Add web compatibility

belgamo:
- Add and fix configs to work fine on web.
- Add patch for react-native-elements slider component


package.json
```diff
@@ -21,6 +21,7 @@
     "expo:test": "yarn workspace @animavita/expo test",
     "graphql": "yarn workspace @animavita/graphql start",
     "graphql:test": "yarn workspace @animavita/graphql test",
+    "postinstall": "yarn workspace @animavita/ui postinstall",
```
renanmav: Make it run with lerna

PR #58, opened 2020-07-19
WIP: Adoptions

### LINKED ISSUE #50: Create adoption mutation
#30

### LINKED ISSUE #51: All adoptions query
#30

belgamo:
Closes #50
Closes #51

### DIFF
--- a/packages/graphql/src/modules/adoption/AdoptionQuery.ts
+++ b/packages/graphql/src/modules/adoption/AdoptionQuery.ts
@@ -1,0 +1,15 @@
+import {GraphQLFieldConfigMap, GraphQLList} from 'graphql';
+
+import {GraphQLContext} from '../../types';
+
+import AdoptionType from './AdoptionType';
+import AdoptionModel from './AdoptionModel';
+
+export const AdoptionsQueries: GraphQLFieldConfigMap<any, GraphQLContext, any> = {
+  adoptions: {
+    type: GraphQLList(AdoptionType),
+    resolve: (obj, args, context) => {
+      return AdoptionModel.find();
+    },
+  },
+};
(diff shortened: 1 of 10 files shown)


packages/graphql/src/modules/adoption/AdoptionQuery.ts
```diff
@@ -0,0 +1,15 @@
+import {GraphQLFieldConfigMap, GraphQLList} from 'graphql';
+
+import {GraphQLContext} from '../../types';
+
+import AdoptionType from './AdoptionType';
+import AdoptionModel from './AdoptionModel';
+
+export const AdoptionsQueries: GraphQLFieldConfigMap<any, GraphQLContext, any> = {
+  adoptions: {
+    type: GraphQLList(AdoptionType),
```
renanmav, line 10: Shouldn’t this be a connection? https://relay.dev/docs/en/graphql-server-specification#connections

How many adoptions this query would return? If is more than 10, it should be a connection. If is not, then leave as it is.

packages/graphql/src/modules/adoption/AdoptionType.ts
```diff
@@ -0,0 +1,107 @@
...
+const AdoptionTypeConfig: ConfigType = {
+  name: 'Adoption',
+  description: 'Animavita adoption',
+  fields: () => ({
+    id: globalIdField('Adoption', adoption => adoption._id),
+    _id: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'MongoDB _id',
+      resolve: user => user._id.toString(),
+    },
```
renanmav, line 55: I don’t think we should expose this

PR #61, opened 2020-12-29
Expo update to SDK 40

petruspierre:
## Summary
Expo SDK 36 have been deprecated since December 9th, so this PR will resolve that. Many packages were updated, so many adjustments had to be made.

## Changelogs
- Expo packages were updated using CLI command `expo upgrade`. Now using [Expo SDK 40](https://blog.expo.io/expo-sdk-40-is-now-available-d4d73e67da33?gi=fa807b0055a5)
- The new packages generated a error on ESLint, so `import/namespace` rule was deactivated and `import/no-cycle` removed from .eslintrc - [Related issue](https://github.com/typescript-eslint/typescript-eslint/issues/1333#issuecomment-573332830)
- Updated testing libraries and jest config and ts config files
- Updated tests snapshots using `jest -u`
- Changed `fireEvent` syntax on Switch component unit test [API Documentation](https://callstack.github.io/react-native-testing-library/docs/api/#fireeventeventname)

PR #63, opened 2021-01-08
Fix bugs

belgamo:
## Summary
Facebook SDK have been updated in #61 and some changes were needed in order to keep it working. One of the things that I need to change was the way how the expires date is stored on database. Before that, the facebook SDK sent the expires date as `int` and now it's returned as a `date`. Besides that, there was a bug on the base64 image parser and the images from UI package didn't appear on android.

## Changelog
- Update facebook SDK syntax
- Get the difference between now and expires date to store it as `int`
- Replace base64 image parser
- Fix android image bug related to [this issue](https://github.com/expo/expo/issues/7545)

### DIFF
--- a/packages/expo/metro.config.js
+++ b/packages/expo/metro.config.js
@@ -1,3 +1,28 @@
 const {createMetroConfiguration} = require('expo-yarn-workspaces');
 
-module.exports = createMetroConfiguration(__dirname);
+const config = createMetroConfiguration(__dirname);
+
+config.server = {
+  enhanceMiddleware: middleware => {
+    return (req, res, next) => {
+      const assets = '/ui/assets/';
+      if (req.url.startsWith(assets)) {
+        // At the beginning of the path to your assets should always be "/assets/"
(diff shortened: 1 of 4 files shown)


BeAmarawn (approved):

BeAmarawn (approved):

petruspierre (approved):

packages/expo/metro.config.js
```diff
@@ -1,3 +1,28 @@
 const {createMetroConfiguration} = require('expo-yarn-workspaces');
 
-module.exports = createMetroConfiguration(__dirname);
+const config = createMetroConfiguration(__dirname);
+
+config.server = {
+  enhanceMiddleware: middleware => {
```
brunohkbx, line 6: What do you think about moving this to a function at the top? would be easier to read.

belgamo: Sure! Will do in the next PR! Thanks!

PR #66, opened 2021-01-14
Feat carrousel swiper

BeAmarawn:
### Summary

- Added a name to android package
- Update on some dependencies related with React Native Reanimated
- Created an interface to represent input user data
- Created a custom Card Component
- Created a native Deck Swiper animation, feeded by any array following the profile's interface, each Card when is swiped has a -- Callback function to remove it from the list and can perform another action.

### Changelog
- Added a android package name on 'packages/expo/app.json'
- Added a mock images on 'packages/ui/assets/images/examples'
- Added an Profile interface on 'packages/ui/@types'
- Added a custom card component on 'packages/ui/core/Card'
- Added a native custom  Deck Swiper animation on 'packages\expo\src\modules\home\Home.tsx'

### Specific modules used
[React Native Reanimated](https://github.com/software-mansion/react-native-reanimated)
[React Native Gesture Handler](https://github.com/software-mansion/react-native-gesture-handler)

Screenshot3
Screenshot2
Screenshot1

https://user-images.githubusercontent.com/38305896/104603715-716e5180-565b-11eb-922f-49dd40aaf8f6.mp4

PR #77, opened 2021-04-26
Add adoption list queries

### LINKED ISSUE #74: Create all adoptions and my adoptions query
## Overview
- A query that returns all adoption options closest to the user.
- A query that returns all adoptions the user has posted.

#30

belgamo:
closes #74 

#### Why I decided to use `forwardConnectionArgs` instead `connectionArgs`?
Because in the app's home screen slider we can't go back to a previous adoption as shown in [this PR](https://github.com/animavita/animavita/pull/66).

#### CHANGELOG
- Added adoption graphql types and model
- Added missing .editorconfig file
- Updated GraphQL schema

The improved version of [this](https://github.com/animavita/animavita/pull/58).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,4 +5,3 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
--- a/packages/graphql/src/modules/adoption/AdoptionModel.ts
+++ b/packages/graphql/src/modules/adoption/AdoptionModel.ts
@@ -2,9 +2,22 @@ import mongoose, {Document} from 'mongoose';
 
 const {ObjectId} = mongoose.Types;
 
-type AnimalGenders = 'male' | 'female';
-type AnimalTypes = 'dog' | 'cat' | 'other';
-type AnimalSizes = 'small' | 'medium' | 'big';
+enum AnimalGenders {
+  male = 'male',
+  female = 'female',
+}
+
+enum AnimalTypes {
+  dog = 'dog',
+  cat = 'cat',
+  other = 'other',
+}
+
+enum AnimalSizes {
+  small = 'small',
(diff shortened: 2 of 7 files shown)


.editorconfig
```diff
@@ -0,0 +1,8 @@
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
```
andrerpbts: Extra line

packages/graphql/src/modules/adoption/AdoptionQuery.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import AdoptModel from './AdoptionModel';
+import AdoptionType from './AdoptionType';
+
+export const AdoptionConnection = connectionDefinitions({
+  name: 'Adoption',
+  nodeType: AdoptionType,
+});
+
+export const AdoptionsQueries: GraphQLFieldConfigMap<any, GraphQLContext, any> = {
```
renanmav, line 14: Why not specifying the first generic type of `GraphQLFieldConfigMap`?

renanmav (commented):
Nice job!

petruspierre:
LGTM 👏

packages/graphql/src/modules/adoption/AdoptionModel.ts
```diff
@@ -0,0 +1,81 @@
+import mongoose, {Document} from 'mongoose';
+
+const {ObjectId} = mongoose.Types;
+
+type AnimalGenders = 'male' | 'female';
```
brunohkbx: Enum works here: https://www.typescriptlang.org/docs/handbook/enums.html

packages/graphql/src/modules/adoption/AdoptionQuery.ts
```diff
@@ -0,0 +1,28 @@
...
+  adoptions: {
+    type: AdoptionConnection.connectionType,
+    args: forwardConnectionArgs,
+    resolve: (_, args) => connectionFromPromisedArray(AdoptModel.find().populate('user') as any, args),
```
brunohkbx: IMO I think it's better to move these resolvers to another file. They can get big over time and will be hard to read having them inline.

.editorconfig
belgamo: Linter default config...

PR #78, opened 2021-04-29
Add i18n package

petruspierre:
Add i18n package using Context API and [i18n-js](https://github.com/fnando/i18n-js).

**DEMO**:

https://user-images.githubusercontent.com/39802870/116563678-9f119a80-a8da-11eb-9a5e-e9bc2989d31f.mp4

In this case, we have support to EN and PT, but not for ES, so the app will show in PT, which is the default language.

### DIFF
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -23,6 +23,7 @@
     "expo": "^40.0.0",
     "expo-constants": "~9.3.3",
     "expo-facebook": "~9.1.0",
+    "expo-localization": "~9.1.0",
     "react": "16.13.1",
     "react-dom": "16.13.1",
     "react-facebook-sdk": "^1.1.1",
--- a/packages/expo/src/navigation/index.tsx
+++ b/packages/expo/src/navigation/index.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import {NavigationContainer} from '@react-navigation/native';
 import {createStackNavigator} from '@react-navigation/stack';
+import * as Localizaiton from 'expo-localization';
 import {ThemeContext} from '@animavita/theme';
+import {useI18nController} from '@animavita/i18n';
 
 import HomeNavigator from './Home';
 import AuthNavigator from './Auth';
@@ -10,6 +12,12 @@ const Stack = createStackNavigator();
 const {Navigator, Screen} = Stack;
 
 export default function Navigation() {
+  const {setLocale} = useI18nController();
+
+  useEffect(() => {
+    setLocale(Localizaiton.locale);
+  }, []);
+
   return (
     <ThemeContext.Consumer>
       {theme => (
--- a/packages/i18n/src/Controller.tsx
+++ b/packages/i18n/src/Controller.tsx
@@ -1,0 +1,41 @@
+import React, {useState, useEffect, useContext, createContext} from 'react';
+import i18n from 'i18n-js';
+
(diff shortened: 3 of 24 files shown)


package.json
```diff
@@ -30,7 +31,7 @@
     "relay:watch": "yarn workspace @animavita/relay watch",
     "start": "yarn workspace @animavita/expo start",
     "storybook": "yarn workspace @animavita/ui storybook",
-    "test": "yarn expo:test && yarn ui:test && yarn graphql:test",
+    "test": "yarn expo:test && yarn ui:test && yarn graphql:test && yarn i18n:test",
```
renanmav: We should refactor this into setting up jest in a monorepo. 

Running `yarn test` in the root should trigger all package's test files under each package's `jest.config.js`

packages/expo/src/navigation/index.tsx
```diff
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import {NavigationContainer} from '@react-navigation/native';
 import {createStackNavigator} from '@react-navigation/stack';
+import * as Localizaiton from 'expo-localization';
```
renanmav: Typo here

packages/i18n/src/Controller.tsx
```diff
@@ -0,0 +1,41 @@
...
+const I18nContext = createContext<I18nContextProps | undefined>(undefined);
+
+const I18nProvider: React.FC = ({children}) => {
+  const [locale, setLocale] = useState<string>('pt-BR');
+
+  i18n.fallbacks = true;
+  i18n.defaultLocale = 'pt-BR';
+
+  useEffect(() => {
+    i18n.locale = locale;
```
renanmav, line 21: Shouldn't we persist this data to handle better communications/email? Like, this should be a user preference.

packages/i18n/src/useI18n.ts
```diff
@@ -0,0 +1,22 @@
+import {useI18nController} from './Controller';
+
+const useI18n = (namespaces: string[]) => {
+  const context = useI18nController();
+
+  const t = (key: string, options: object = {}) => {
```
renanmav: Shouldn't we type this `options` better? It uses `context.t` second argument type.

packages/i18n/src/Controller.tsx
petruspierre: So what do you recommend that we do?

packages/expo/package.json
```diff
@@ -23,6 +23,7 @@
     "expo": "^40.0.0",
     "expo-constants": "~9.3.3",
     "expo-facebook": "~9.1.0",
+    "expo-localization": "~9.1.0",
```
belgamo, line 27: I recommend you add the i18n package as dependency here as well. We're gonna update all the package.json files adding the respectives dependencies in order to keep the consistence among them.

belgamo (approved):
Nice job!

packages/i18n/src/Controller.tsx
belgamo: Maybe create a screen accessible by configuration menu where the user can switch the app language and persist this info on the backend side. What do you think about doing this in another task/pr @renanmav?

belgamo: #80 and #81

packages/expo/package.json
belgamo: #82

PR #84, opened 2021-05-12
Dockerize graphql package externals services

belgamo:
### Summary
The goal is to make easier for the new contributors the setup of this project and decrease the need to create externals accounts like AWS which requires a credit card.

### Changelog
- Added a docker-compose file in GraphQL package
- Updated contribution guide
- Updated env vars config files to replace them when using docker

### DIFF
--- a/packages/graphql/docker-compose.yml
+++ b/packages/graphql/docker-compose.yml
@@ -1,0 +1,27 @@
+version: '2.1'
+
+services:
+  mongo:
+    container_name: animavita-mongo
+    image: mongo
+    restart: always
+    ports:
+      - 27017:27017
+  localstack:
+    container_name: animavita-localstack
+    image: localstack/localstack
+    network_mode: bridge
+    ports:
+      - '4566:4566'
+      - '${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}'
+    environment:
+      - SERVICES=s3,sqs
+      - DEBUG=1
+      - DATA_DIR=/tmp/localstack/data
+      - PORT_WEB_UI=${PORT_WEB_UI- }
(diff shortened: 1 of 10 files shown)


packages/graphql/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: '2.1'
+
+services:
+  mongo:
+    container_name: animavita-mongo
+    image: mongo
```
brunohkbx: I think it would be better to specify the version you are running it. just to avoid breaking changes

packages/graphql/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    restart: always
+    ports:
+      - 27017:27017
+  localstack:
+    container_name: animavita-localstack
+    image: localstack/localstack
```
brunohkbx: same here

petruspierre:
Great job 🚀

petruspierre:
I think it's valid to investigate if Android will have trouble finding the bucket data, now that it's on localhost. 
For example, profile pictures may not appear in the android app.

hjJunior (approved):

packages/graphql/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    network_mode: bridge
+    ports:
+      - '4566:4566'
+      - '${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}'
+    environment:
```
hjJunior, line 17: Just thinking out loud, should we consider having .env for these envs?

belgamo: I don't know... Thinking about docker-compose file, if someone needs to change some env it can be made directly here.

PR #89, opened 2021-05-24
Fix bugs on web version

belgamo:
### Summary
This project also works on web. But when I tried to run `yarn web` I faced some problems. Basically, webpack wasn't configured properly to handle some packages that were added in previous features. Another problem is about `expo-facebook` because [it doesn't work](https://github.com/expo/expo/pull/6862) on web.

### Solution
To solve webpack problems, I just include all missing modules in webpack config. About the facebook problem, I had to create a specific component for web in order to fix that opening a modal provided by `react-facebook-login`. To avoid duplicate code in both login components, I created a custom hook to share common code.

### Changelog
- Updated enviroments examples to avoid issues on setup
- Fixed localstack issues on MacOS
- Fixed web project

### To do
#90

### DIFF
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -30,7 +30,8 @@
     "expo-localization": "~9.1.0",
     "react": "16.13.1",
     "react-dom": "16.13.1",
-    "react-facebook-sdk": "^1.1.1",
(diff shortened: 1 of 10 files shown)


petruspierre (approved):
Great job! 🚀

PR #93, opened 2021-06-14
Refactor backend architecture

Beigelman:
## Reason
This PR has the goal to uncouple various layers of our backend architecture. This is still a work in progress but with one or two other PRs this should be done
The new architecture proposal is the following one:

1. Domain: Responsible for the app domain and all its business logic
2. Infra: In charge of all the communication with external API and Databases
3. App: Control all the use cases for each different domain
4. Presentation: The layer responsible to communicate with all the client in our case using GraphQl to serve our frontend

The test strategy was also changed to suit the new architecture.

## Screenshots

image
image
image


packages/graphql/jest.config.js
```diff
@@ -1,20 +1,180 @@
-const pkg = require('./package');
+// For a detailed explanation regarding each configuration property, visit:
```
belgamo, line 1: Could you clear this file by removing all comments?

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
belgamo: I think it's a user settings level. Can you add it to `.gitignore`?

Beigelman: Done!

packages/graphql/jest.config.js
Beigelman: Sure

packages/graphql/jest.config.js
```diff
@@ -1,20 +1,13 @@
...
+// https://jestjs.io/docs/en/configuration.html
 
 module.exports = {
-  rootDir: './',
```
brunohkbx, line 4: Are these changes really necessary?

packages/graphql/package.json
```diff
@@ -4,19 +4,22 @@
...
+    "start:infra": "docker-compose up -d",
     "test": "jest",
     "test:watch": "jest --watch",
-    "update:schema": "babel-node --extensions \".es6,.js,.es,.jsx,.mjs,.ts,.tsx\" ./scripts/updateSchema.ts"
+    "update:schema": "ts-node-dev ./scripts/updateSchema.ts"
   },
   "dependencies": {
     "@entria/graphql-mongoose-loader": "^4.3.0",
     "@koa/router": "^8.0.6",
+    "awilix": "^4.3.4",
```
brunohkbx, line 19: Why these packages were added?

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+    authenticateUser,
+    storageProvider,
+    userRepository,
+    facebookRepository,
+    tokenProvider,
+  };
+};
+
+describe('AuthenticateFacebookUser', () => {
+  it('should be able to authenticate', async () => {
```
brunohkbx: Not a big thing but tests shouldn't start with `should` they need to be assertive about what they are doing
https://www.betterspecs.org/#should

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+        {
+          id: 'socialId-fake-token',
+          providedBy: 'facebook',
+        },
+      ],
+    });
+    expect(token).toMatch(/(JWT )\w+\.\w+\.\w+/g);
+  });
+
+  it('should be able to authenticate with no user photo', async () => {
```
brunohkbx: Can we break down these different contexts in describe blocks?
Something like:
```js
describe('When the user has no photo', () => {
  it('authenticates it, () => {
   ....
  });
})

Comment (brunohkbx) [reply packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts]: The same for all others below

File: packages/graphql/src/modules/user/app/authenticateFacebookUser.ts [hunk 7]
```diff
@@ -0,0 +1,123 @@
...
+  expires: number;
+  permissions: string[];
+}
+
+export interface SaveFBResponse {
+  user: User;
+  token: string;
+}
+
+const authenticateFacebookUser = ({
```
brunohkbx, line 26: @Beigelman btw shouldn't this be part of a different PR? I mean, the point here is to refactor the backend architecture right?

brunohkbx:
Nice work here @Beigelman. However, I suggest breaking down the new things in separated PRs in order to make them easier to review. This PR should include only changes regarding the new architecture
Architecture-wise, I wonder if wouldn't be better to stick with the current organization. I mean, this is an open-source project meant to receive help from other people. Applying DDD here definitely increases the learning curve to board new people in the project. Up to you @belgamo

packages/graphql/package.json
Beigelman: Awilix is an injection dependency package that helps decouple several layers of the application. It is a great tool for implements some SOLID concepts.

packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
Beigelman: The previous implementation was basically doing everything in one file, since I've isolated different parts of the application I had to refactor this use case as well.

packages/graphql/jest.config.js
Beigelman: Yes, due to the nature of the new architecture proposal there were I few things that needed to be changed in order to adjust to this architecture.

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
Beigelman: Cool, I didn't know that I will address this issue

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
Beigelman: That's a good suggestion, thanks!

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
Beigelman: Pretty cool this site https://www.betterspecs.org/ by the way!

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
Beigelman: Okay 👍🏼

Beigelman:
> Nice work here @Beigelman. However, I suggest breaking down the new things in separated PRs in order to make them easier to review. This PR should include only changes regarding the new architecture
> Architecture-wise, I wonder if wouldn't be better to stick with the current organization. I mean, this is an open-source project meant to receive help from other people. Applying DDD here definitely increases the learning curve to board new people in the project. Up to you @belgamo

I agree that this could have been broken down into separated PRs and is something that I will have in mind in my next contributions.
Architecture-wise I disagree with you. I find the current architecture very unorganized and difficult to understand. My intention with this PR is to turn the code more readable and easy to contribute.

Thanks for all your comments I will address them soon.

belgamo:
> Nice work here @Beigelman. However, I suggest breaking down the new things in separated PRs in order to make them easier to review. This PR should include only changes regarding the new architecture
> Architecture-wise, I wonder if wouldn't be better to stick with the current organization. I mean, this is an open-source project meant to receive help from other people. Applying DDD here definitely increases the learning curve to board new people in the project. Up to you @belgamo

Nice job @Beigelman and thanks for your review @brunohkbx, I really appreciate it! 

@brunohkbx since that I've started to contribute here I feel the things a little unorganized on the backend side. I mean, a lot of different routines placed together, one class with various responsibilities. Then in this point, I agree with @Beigelman about applying any concept which goal is to improve the organization of the code. 
About the fact it is an opensource project I agree we need to keep things simple as much as possible. But I prefer the architecture a little bit harder to understand over a bad codebase. That's worthwhile.

belgamo (approved):
Nice job! You just forgot about integration tests and to applying DDD to the adoption concept. Since this PR is big enough, I will map these tasks to be done in another one.

PR #94, opened 2021-06-15
fix: fixes s3 missing objects

### LINKED ISSUE #87: S3 objects aren't appering on android
### Problem
After #84 the images stopped appering on android. That happens only when we run graphql package in docker. So, I suppose the app are trying to get the images with "localhost" in URL which doesn't work on Android. Probably because when using docker, we mock all AWS dependencies locally such as S3 that no longer stores a public object URL which was accesible to both (android/ios) in the past.

### Proposed solution
I think changing URL of S3 objects to points to `10.0.2.2` will fix this problem.

To testing, the user picture must appear in the header.
image

### Reference
https://developer.android.com/studio/run/emulator-networking

efgalvao:
Fixes #87 

Fixes the S3 missing objects in android devices while using localstack by getting the machine ip set in configuration to create the correct path to load objects. Adds a `--https` to expo web so the user can access facebook correctly while in web version.

* Sample:

Screenshot_1623763048

### DIFF
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -12,7 +12,7 @@
     "start": "expo start",
     "test": "jest",
     "test:watch": "jest --watch",
-    "web": "expo start --web"
+    "web": "expo start --web --https"
   },
   "dependencies": {
     "@animavita/i18n": "1.0.0",
@@ -42,6 +42,7 @@
     "styled-components": "5.0.0"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.14.2",
     "@expo/webpack-config": "~0.12.45",
     "@testing-library/jest-native": "^3.4.3",
     "@testing-library/react-native": "^7.1.0",
@@ -51,7 +52,6 @@
     "expo-yarn-workspaces": "^1.2.1",
     "jest": "^26.6.3",
     "jest-expo": "^40.0.0",
-    "jest-styled-components": "^7.0.3",
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.14.2"
+    "jest-styled-components": "^7.0.3"
   }
 }
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -38,6 +38,7 @@
     "react-native-reanimated": "~1.13.0",
     "react-native-safe-area-context": "3.1.9",
     "react-native-screens": "~2.15.0",
+    "react-native-url-polyfill": "^1.3.0",
     "react-native-web": "0.13.12",
     "styled-components": "5.0.0"
   },
--- a/packages/expo/src/navigation/Home.tsx
+++ b/packages/expo/src/navigation/Home.tsx
@@ -11,6 +11,7 @@ import Menu, {MenuItem, MenuDivider} from 'react-native-material-menu';
 import {useI18n} from '@animavita/i18n';
 
 import Home from '../modules/home/Home';
+import getEnvVars from '../../../relay/environment';
 
 import {HomeQuery} from './__generated__/HomeQuery.graphql';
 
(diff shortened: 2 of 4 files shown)


packages/expo/src/navigation/Home.tsx
```diff
@@ -52,6 +53,10 @@ const HomeNavigator: React.FC = () => {
   const name = me?.name?.split(' ')[0] || '';
   const imageIndex = me ? me.profileImages.length - 1 : 0;
   const uri = me?.profileImages[imageIndex].url;
+  const uriend = uri.slice(16);
```
belgamo: Nice job!

I don't think this is a proper way to work with URL because the URL may changes in the future which makes your code break. Instead, I prefer methods where you use patterns to get specific parts of the link like regex or some library like [react-native-url-polyfill](https://github.com/charpeni/react-native-url-polyfill).

efgalvao: I have used the library you suggested and made, thank you for the hint.

packages/expo/package.json
```diff
@@ -42,6 +42,7 @@
     "styled-components": "5.0.0"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.14.2",
```
brunohkbx, line 46: So instead of adding separated plugins, what do you think about using the preset from expo? 
Check it out: https://github.com/expo/expo/tree/master/packages/babel-preset-expo

It includes `@babel/preset-env` which already includes most plugins out there plus other configs like lazyImports
cc @belgamo

belgamo: LGTM. I would do that in another PR.

packages/ui/package.json
```diff
@@ -23,6 +23,7 @@
     "react-native-ratings": "^7.2.0",
     "react-native-reanimated": "~1.13.0",
     "react-native-svg": "^11.0.1",
+    "react-native-url-polyfill": "^1.3.0",
```
belgamo: It shouldn't be here. Please put this in the package.json of expo package.

packages/expo/src/navigation/Home.tsx
efgalvao: I think this lib is causing an error in web mode. Can you test ?

efgalvao:
@belgamo , I think the react-native-url-polyfill lib is causing a crash in web mode, can you check if is causing in you system too ?

belgamo:
> @belgamo , I think the react-native-url-polyfill lib is causing a crash in web mode, can you check if is causing in you system too ?

Nice @efgalvao. I've tried to run it and a got an error.

image

To solve that I just added `react-native` in the list of node dependencies in the webpack config so that webpack can handle it. Probably because `react-native-url-polyfill` uses something from `react-native`.

belgamo (approved):

charpeni:
I'm the author of `react-native-url-polyfill` 👋 

Thanks for documenting this @belgamo. It does use something from `react-native` (`polyfillGlobal`) in order to polyfill.

I'm curious if we really need to apply those polyfills if it's running in web mode? Should we do something different to improve the developer experience here? Maybe we could check the platform and do it without `polyfillGlobal` on web.

belgamo:
> I'm the author of `react-native-url-polyfill` 
> 
> Thanks for documenting this @belgamo. It does use something from `react-native` (`polyfillGlobal`) in order to polyfill.
> 
> I'm curious if we really need to apply those polyfills if it's running in web mode? Should we do something different to improve the developer experience here? Maybe we could check the platform and do it without `polyfillGlobal` on web.

Hi @charpeni, thanks for your help here!

Yeah, I think checking the platform would be a nice way to improve the developer experience. Let us know when you do that, thanks a lot!

PR #100, opened 2021-07-14
Setup integration tests

### LINKED ISSUE #96: Add integration tests

Beigelman:
closes #96

Add integration test for `authenticateFacebookUser`


.nvmrc
```diff
@@ -0,0 +1 @@
+v14.17.0
```
belgamo, line 1: Add this file to your `.gitignore`.

packages/graphql/src/modules/user/presentation/controllers/facebookSessionController.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      const {
+        authenticateFacebookUser: {user, token, error},
+      } = response.body.data;
+
+      expect(user.name).toBe('John Doe');
+      expect(token).toMatch(/(JWT )\w+\.\w+\.\w+/g);
+      expect(error).toBeNull();
+    });
+  });
```
belgamo, line 74: You should test the opposite case as well.

belgamo (changes requested):

.nvmrc
Beigelman: But the point of this file is to show other contributors what version of node to use and help them install the correct version by running `nvm install` or `nvm use`

packages/graphql/src/modules/user/presentation/controllers/facebookSessionController.spec.ts
Beigelman: I did that at first but I don't see the point since I already test the failure case in the unit test. Another thing that I notice is that it will console log the [errors](https://github.com/animavita/animavita/pull/100/files#diff-12f7960f82066c8f974f9f1c82dc1851872f5c34e036dbebea76743ebc3737ecR113) during the run. I'm not sure if we want that to happen.

.nvmrc
belgamo: Oh nice! That's ok.

packages/graphql/src/modules/user/presentation/controllers/facebookSessionController.spec.ts
```diff
@@ -0,0 +1,66 @@
+import * as http from 'http';
+
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+import request from 'supertest';
+
+import app from '../../../../app';
+import {connectMongoose, clearDbAndRestartCounters, disconnectMongoose} from '../../../../shared/tests/helpers';
```
brunohkbx, line 8: what about having some aliases here? I think you can do that in babel

packages/graphql/src/modules/user/presentation/controllers/facebookSessionController.spec.ts
brunohkbx: You can mock console.error to disable logging at the console. I don't see any problem repeating the test again. Actually I think it's great because even though there's a unit test, having an integration test will significantly increase the confidence around this flow.

brunohkbx: You can mock console.error to disable logging at the console. I don't see any problem repeating the test again. Actually I think it's great because even though there's a unit test, having an integration test will significantly increase the confidence around this flow.

Beigelman: Done ✅

packages/graphql/src/modules/user/presentation/controllers/facebookSessionController.spec.ts
Beigelman: Do you mean to avoid the ../../.. ...?

belgamo (approved):
Nice PR! I tried to run `yarn install` and I got an error from yarn about the version in use. That's ok, but could you add this requirement to the README.md?

PR #101, opened 2021-07-29
Refactor loaders

Beigelman:
This PR closes a cycle of a series of three PRs that aim to refactor the user module to focus more on the domain of the application. The following changes were made in this PR:
- Refactor the userRepository previous created to deal only with persistency;
- Insert loader in the presentation layer;
- Minor schemas updates;


packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
```diff
@@ -49,7 +49,8 @@ const authenticateFacebookUser = ({
   if (existingUser) {
     // verify if name was updated
     if (name.length !== existingUser.name.length) {
-      await userRepository.updateUser({id: existingUser.id, proprieties: {name}});
+      existingUser.name = name;
```
belgamo: I think in this case it's not a good practice to change an object directly. Is there a chance this causes side effects? How about creating a copy of this object spreading the existing properties and changing the name?

```
const targetUser = { ...existingUser, name };
await userRepository.update(targetUser)
```

packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
```diff
@@ -59,10 +60,9 @@ const authenticateFacebookUser = ({
         providerId.providedBy === 'facebook' ? ({id: fbID, providedBy: 'facebook'} as ProvidersId) : providerId,
       );
 
-      await userRepository.updateUser({
-        id: existingUser.id,
-        proprieties: {providersIds: updatedProvidersIds},
-      });
+      existingUser.providersIds = updatedProvidersIds;
```
belgamo: Same here. If you have this kind of assignment in other places of your code, would be great to improve that.

packages/graphql/src/modules/user/presentation/controllers/userController.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+            variables: {id: userId},
+          });
+
+        const {user} = response.body.data;
+
+        expect(user.name).toBe('John Doe');
+        expect(user.providerIds[0]).toEqual({id: 'fake-id', providedBy: 'facebook'});
+        expect(user.emails[0]).toEqual({email: 'johndoe@animavita.com', providedBy: 'facebook'});
+      });
+    });
```
belgamo, line 144: Break line here :nail_care:

belgamo (changes requested):

packages/graphql/src/interfaces/NodeInterface.ts
```diff
@@ -15,21 +11,7 @@ export function registerType(type: GraphQLObjectType) {
...
-  [key: string]: Loader;
-};
-
 export const {nodeField, nodeInterface: NodeInterface} = nodeDefinitions(
-  (globalId, context: GraphQLContext) => {
-    const {type, id} = fromGlobalId(globalId);
-    const loader: Loader = ((loaders as unknown) as Loaders)[`${type}Loader`];
-
-    return (loader && loader.load(context, id)) || null;
-  },
```
renanmav, line 33: Why are you removing this loading function?

It would make Relay unable to fetch a Node object.

Please provide a way for Relay to map from an object to the GraphQL type associated with that object.

packages/graphql/src/app.ts
```diff
@@ -44,7 +42,7 @@ if (process.env.NODE_ENV !== 'test') {
 app.use(convert(cors({maxAge: 86400, origin: '*'})));
 
 app.use(async (ctx, next) => {
-  ctx.dataloaders = getDataLoaders(loaders);
+  ctx.container = container.cradle;
```
renanmav, line 45: Why are you removing `dataloaders` from Koa context?

packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
```diff
@@ -47,43 +47,37 @@ const authenticateFacebookUser = ({
   const existingUser = users && users[0];
 
   if (existingUser) {
+    let updatedUser = existingUser;
```
belgamo, line 50: You don't need to create this var. Just call update function passing a new object: `await userRepository.update({...existingUser, name});`. Try to avoid assignments.

packages/graphql/src/modules/user/presentation/routes/queries/UserLoader.ts
```diff
@@ -1,51 +0,0 @@
-import DataLoader from 'dataloader';
```
belgamo, line 1: Why this file was deleted? Loader is important for performance stuff.

belgamo (changes requested):

packages/graphql/src/app.ts
Beigelman: `dataloaders` are database related so I plan on moving them to their respective repositories

packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
Beigelman: How would I return the updated user? Do you suggest querying it from the database? I

belgamo: Hum sorry, I didn't notice that you're using this var in other routines after update the user. Let it be.

packages/graphql/src/modules/user/infra/mongoose/repositories/mongoUsersRepository.ts
```diff
@@ -1,10 +1,13 @@
+import DataLoader from 'dataloader';
 import {v4 as uuid} from 'uuid';
 
 import User from '../../../domain/User';
 import UsersRepository, {FindUserByEmailOrProviderId} from '../../../domain/UsersRepository';
 import UserModel, {IUserDocument} from '../models/UserModel';
 
 export default function mongoUsersRepository(): UsersRepository {
+  const userLoader = new DataLoader<string, User>(userIds => UserModel.find({id: {$in: userIds}}));
```
belgamo, line 9: Cool, I think it's enough. But we had a `UserLoader` class with some methods. I don't know how useful they are,  @renanmav could you clarify? [UserLoader](https://github.com/animavita/animavita/blob/next/packages/graphql/src/modules/user/presentation/routes/queries/UserLoader.ts)

packages/graphql/src/app.ts
renanmav: Which repository?

belgamo: DDD Repository layer.

packages/graphql/src/modules/user/infra/mongoose/repositories/mongoUsersRepository.ts
Beigelman: As far as I know, those methods weren't used anywhere. Maybe @renanmav had some plans for them but I don't think they were being used.

packages/graphql/src/app.ts
renanmav: @belgamo could you provide some links about that architecture? I'd like to know some more.

packages/graphql/src/modules/user/infra/mongoose/repositories/mongoUsersRepository.ts
renanmav: Yeah, I don't think they are being used

packages/graphql/src/interfaces/NodeInterface.ts
belgamo: Sure.

[DDD_Reference_2015-03.pdf](https://github.com/animavita/animavita/files/6954903/DDD_Reference_2015-03.pdf)
[DomainDrivenDesignQuicklyOnline (1).pdf](https://github.com/animavita/animavita/files/6954904/DomainDrivenDesignQuicklyOnline.1.pdf)

PR #102, opened 2021-08-09
Refactor user domain to a functional paradigm

Beigelman:
This PR proposes to transfer more of the business logic into the domain layer while applying a functional paradigm to the layer that way, possibly avoiding concurrent problems.

### DIFF
--- a/packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
+++ b/packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
@@ -49,13 +49,20 @@ const authenticateFacebookUser = ({
 
     updatedUser = User.updateFacebookId(updatedUser, fbID);
 
-    if (User.shouldUpdateFacebookProfileImage(updatedUser) && profileUrl) {
+    const existingFbProfileImages = updatedUser.profileImages.find(
+      profileImage => profileImage.providedBy === 'facebook',
+    );
+    const shouldUpdateProfileImage = !existingFbProfileImages || updatedUser.profileImages.length === 0;
+
+    if (shouldUpdateProfileImage && profileUrl) {
       const url = await storageProvider.saveFile({userId: updatedUser.id, imageURL: profileUrl});
 
       updatedUser = User.updateFacebookProfileImage(updatedUser, url);
     }
 
-    await userRepository.update(updatedUser);
+    if (JSON.stringify(user) !== JSON.stringify(updatedUser)) {
+      await userRepository.update(updatedUser);
+    }
 
     return {
       user: updatedUser,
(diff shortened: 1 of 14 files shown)


packages/graphql/src/modules/user/app/authenticateFacebookUser.ts
```diff
@@ -39,66 +39,37 @@ const authenticateFacebookUser = ({
...
-
-      updatedUser = {...updatedUser, providersIds: updatedProvidersIds};
-
-      await userRepository.update(updatedUser);
-    }
-
-    const existingFbProfileImages = updatedUser.profileImages.find(
-      profileImage => profileImage.providedBy === 'facebook',
-    );
+    updatedUser = User.updateFacebookId(updatedUser, fbID);
```
belgamo, line 50: To be honest, I'm wondering about the need for these two functions `updateName` and `updateFacebookId`. Do we really need them? I mean, every time a user logs into we will reach line 58, even when he doesn't have any data to be updated. So, in order to make it simpler, I would create a `shouldUpdateUser` which receives a user instance in order to avoid a possible database hit. This function should check wheater the user name, facebookId or facebook profile image needs to be updated and return a boolean. 
Just make sure to keep the check and the routine to update facebook profile image since it hits an external provider.

Beigelman: The problem may be the function's name. These two functions actually verify if the desired property needs to be changed

belgamo: I don't think so. Just create a function with a generic name like `shouldUpdateUser`. In the future, we might have more properties to be checked, then it's bad if we would need to create one function for each property in order to update the user.

packages/graphql/src/modules/user/domain/User.ts
```diff
@@ -1,49 +1,95 @@
-export type Providers = 'facebook' | 'google' | 'apple';
+namespace User {
```
belgamo, line 1: How about just export a user constant instead use namespace? I think this kind of syntax is outdated.

packages/graphql/src/modules/user/infra/mongoose/models/UserModel.ts
```diff
@@ -57,7 +57,7 @@ const ProfileImageSchema = new mongoose.Schema(
   },
 );
 
-export type IUserDocument = User & Document;
+export type IUserDocument = User.Type & Document;
```
belgamo, line 60: Why do we need a specific `Type` property? Is there some advantage in this approach? Declaring just with `User` is simpler.

packages/graphql/src/modules/user/domain/User.ts
```diff
@@ -1,49 +1,95 @@
...
+  type Providers = 'facebook' | 'google' | 'apple';
 
-export interface ProvidedBy {
-  providedBy: Providers;
-}
+  type ProvidedBy = Readonly<{
```
belgamo, line 4: Do we really need `ReadOnly`? If we focus on functional programming I think using `const` is enough to avoid mutability problems. I suggest turn on some kind of eslint rule so that we can alert reassignments in `let` vars.

belgamo:
This will continue in https://github.com/animavita/animavita/pull/105

PR #105, opened 2021-08-24
Refactor user

Beigelman:
## Overview
This PR refactor the user provider's information inside the user's model aggregating everything related to the provider inside a provider field.

### DIFF
--- a/packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
+++ b/packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
@@ -49,14 +49,23 @@ const setUp = () => {
 };
 
 describe('AuthenticateFacebookUser', () => {
+  const RealNow = Date.now;
+
+  beforeAll(() => {
+    global.Date.now = jest.fn(() => 0);
+  });
+
+  afterAll(() => {
+    global.Date.now = RealNow;
+  });
+
   describe('When logging in for the first time', () => {
     it('authenticates', async () => {
-      jest.spyOn(Date, 'now').mockImplementation(() => 0);
       const {authenticateUser} = setUp();
 
       const {user, token} = await authenticateUser(authenticateUserParams);
 
-      expect(user).toEqual(fakeUser);
+      expect(user).toStrictEqual(fakeUser);
       expect(token).toMatch(JwtRegex);
     });
 
@@ -99,7 +108,7 @@ describe('AuthenticateFacebookUser', () => {
 
       const {user} = await authenticateUser(authenticateUserParams);
 
-      expect(user).toEqual(fakeUser);
+      expect(user).toStrictEqual(fakeUser);
     });
   });
 
--- a/packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
+++ b/packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
@@ -1,8 +1,9 @@
 import User from '../domain/User';
 import createFakeStorageProvider from '../../../shared/providers/StorageProvider/fakes/fakeStorageProvider';
 import createFakeUsersRepository from '../infra/fakes/fakeUsersRepository';
-import createFakeSocialMediaRepository from '../infra/fakes/fakeSocialMediaRepository';
+import createFakeSocialMediaService from '../infra/fakes/fakeSocialMediaService';
 import createFakeTokenProvider from '../providers/TokenProvider/fakes/fakeTokenProvider';
+import Provider from '../domain/Provider';
 
 import authenticateFacebookUser from './authenticateFacebookUser';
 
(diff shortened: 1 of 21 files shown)


packages/graphql/src/modules/user/infra/fakes/fakeSocialMediaService.ts
```diff
@@ -1,6 +1,6 @@
-import SocialMediaRepository, {SocialUser} from '../../domain/SocialMediaRepository';
+import SocialMediaService, {SocialUser} from '../../services/SocialMediaService';
 
-export default function fakeSocialMediaRepository(): SocialMediaRepository {
+export default function fakeSocialMediaService(): SocialMediaService {
```
belgamo, line 3: It seems more like a mock. What do you think about placing it together with spec files?

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
```diff
@@ -52,68 +43,35 @@ const setUp = () => {
...
-    facebookRepository,
+    facebookService,
     tokenProvider,
   };
 };
 
 describe('AuthenticateFacebookUser', () => {
   describe('When logging in for the first time', () => {
     it('authenticates', async () => {
+      jest.spyOn(Date, 'now').mockImplementation(() => 0);
```
belgamo: I've never seen someone mocking Date like that. I think `spyOn` method is useful when you need to track your mock calls. Otherwise, the "popular" way to do that is to use `jest.fn`. But it's not a big deal, just a suggestion.

https://www.maxpou.fr/jest-mock-date

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
```diff
@@ -52,68 +43,35 @@ const setUp = () => {
...
-          },
-        ],
-        providersIds: [
-          {
-            id: 'socialId-fake-token',
-            providedBy: 'facebook',
-          },
-        ],
-      });
+      expect(user).toEqual(fakeUser);
```
belgamo: How about a deep comparison here?

belgamo:
Nice job @Beigelman! Let's apply the same DDD concepts to the other modules of animavita. The next one is `adoptions` which I've started a few months ago.

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
Beigelman: Do you mean to use deepEqual?

belgamo: I'm used to using `toStrictEqual`.

packages/graphql/src/modules/user/infra/fakes/fakeSocialMediaService.ts
Beigelman: You are right this is a mock. The problem here is that the spec files are placed along with their implementation and this mock might be used by more than one spec file. So, I think letting it in the infra folder might be the best option by now.

Beigelman: Do you think renaming the file to MockedSocialMediaService would describe better the file function?

packages/graphql/src/modules/user/app/authenticateFacebookUser.spec.ts
Beigelman: Cool, this was the way I found to mock it but I think using jest.fn might be better. I will try out

packages/graphql/src/modules/user/infra/fakes/fakeSocialMediaService.ts
belgamo: The name looks good to me. The point is that fake/mocks functions are only used by test files. So, I think it's properly placing them together.

belgamo:
> Nice job @Beigelman! Let's apply the same DDD concepts to the other modules of animavita. The next one is `adoptions` which I've started a few months ago.

@Beigelman please rebase your branch to run tests again and fix the checks.

packages/graphql/src/modules/user/utils/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export const transformDomainUserIntoPresentationUser = (user: User.Type): UserGraphQLType => {
+  const lastLoggedInProvider = user.providers.sort(provider => provider.lastLogIn)[0];
+
+  return {
+    id: user.id,
+    name: lastLoggedInProvider.name,
+    email: lastLoggedInProvider.email,
+    profileImage: lastLoggedInProvider.profileImage,
+  };
+};
```
renanmav: Why is this function located within a `index` file within an empty folder (`utils`)? 

Wouldn't make more sense to name it `utils.ts` and place it within `packages/graphql/src/modules/user`?

PR #121, opened 2022-10-28
Setup database

belgamo:
### Summary
The goal of this is to set up mongo as our database. Also added a docker-compose file to make it easier to start the project.


.gitignore
```diff
@@ -12,3 +13,32 @@ web-build/
...
+*.launch
+.settings/
+*.sublime-workspace
+
+# IDE - VSCode
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
```
mauricio-prl, line 44: EOF line?

m-pereira (approved):
LGTM

abarbosa-godaddy:
LGTM

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.9"
+
+services:
+  mongodb:
+    image: mongo
```
brunohkbx: Wouldn't be better to pin a version to avoid breaking changes? Then we can also have control over the dev environment  and update it accordingly

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+    ports:
+      - 27017:27017
+    environment:
+      - MONGO_INITDB_ROOT_USERNAME=mongo
+      - MONGO_INITDB_ROOT_PASSWORD=mongo
+      - MONGO_INITDB_DATABASE=animavita
```
brunohkbx: Would be nice to also persist the data in a volume otherwise it would (correct me if i'm wrong) be deleted every time you destroy the container

```suggestion
version: "3.9"

services:
  mongodb:
    image: mongo
    ports:
      - 27017:27017
    environment:
      - MONGO_INITDB_ROOT_USERNAME=mongo
      - MONGO_INITDB_ROOT_PASSWORD=mongo
      - MONGO_INITDB_DATABASE=animavita
   volumes:
     - mongo:/data/db
volumes:
  mongo:
```

belgamo: Definitely, thanks!

docker-compose.yml
belgamo: For sure!

PR #122, opened 2022-10-28
Add a base structure for adoption module

belgamo:
Continuing #121 

### Summary
This PR is to add the core module of our app: adoptions. This is actually a POC for our v2 and I want to implement something on the mobile side integrating it with our backend to see how it goes.

### Validation
On NestJS we have a couple of validation strategies: [schema-based](https://docs.nestjs.com/pipes#schema-based-validation) and [decorator-based](https://docs.nestjs.com/pipes#class-validator). I decided to go with the schema-based one because then we can share the schema across our apps. That's I created the `schema-validations` under the shared folder. Also, I'm using [Joi](https://github.com/hapijs/joi) to validate. We have a custom pipe on the backend that integrates Nest with the Joi schemas.

### `schema-validations` packages
So as I said above I added this new package to share schemas. I had to set up typescript properly to generate a dist folder with the transpiled code. Due to this, whenever we make a change on `schema-validations` we need to rebuild this package so the backend app can have access to the new code. To rebuild, we just run: `pnpm -r run build`. Not sure if we can automate this task by watching the files in the workspace. Maybe `Nx` can help with this, but let's do that in the future.


shared/validation-schemas/src/adoption.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  type: Joi.string().max(5),
+
+  age: Joi.number(),
+
+  size: Joi.string().max(6),
+
+  observations: Joi.string().allow("").max(100),
+
+  photos: Joi.array().optional().items(Joi.string()),
```
abarbosa-godaddy: These empty lines. Should we get rid of them?

abarbosa-godaddy:
Worth it to add unit tests to validators and controllers at least?

shared/validation-schemas/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@animavita/validation-schemas",
```
brunohkbx, line 2: Why this need to be shared?

shared/validation-schemas/src/adoption.ts
brunohkbx: Is it possible to use `Joi.string().uri()` instead?

belgamo: I think so, let's try.

shared/validation-schemas/package.json
belgamo: Cause then we can validate on the client as well matching the same schema.

shared/validation-schemas/src/adoption.ts
belgamo: I like this style cause we might have to nest more Joi functions.

belgamo:
> Worth it to add unit tests to validators and controllers at least?

Done!

gquaresma-godaddy (approved):

PR #134, opened 2022-11-12
Configure tests and continuous deployment

belgamo:
## Summary
Basically, the goal of this PR is to set up an action to run our API e2e tests on each opened PR and dockerize our backend. 

## Why dockerize?
For two reasons:
1. I integrated that into the docker-compose. So now you're able to get our infra and backend by running one command. Just run `docker-compose up -d` and you're all set.
2. With a Dockerfile we're ready to deploy to [fly.io](https://fly.io/) or any other service that allows deployments using Dockerfile.

## Issues
This was not trivial due to the fact we're in a monorepo. Thanks to https://github.com/pnpm/pnpm/issues/3114 I was able to get an idea of how things work and what we have to do. [Their docs](https://pnpm.io/cli/fetch) were useful as well. 

### The hoist issue
So basically we're using the `hoisted` strategy for the [node-linker](https://pnpm.io/npmrc#node-linker) config. [Expo needs that](https://docs.expo.dev/guides/monorepos/) to be able to run in a monorepo. However, I started getting a weird error when trying to build the nest application because of that and I'm not really sure why it breaks: 

Screen Shot 2022-11-13 at 23 34 05

It happens either for a docker image building or when running on [GitHub actions](https://github.com/animavita/animavita/actions/runs/3457864651/jobs/5771716582). Setting `node-linker` to `isolated` solves that but then it breaks expo. So my workaround for that was to create a second `.npmrc` with the `node-linker` set to `isolated` and use this file only for backend builds instead of using the default `.npmrc` file.

### The `pnpm deploy` issue
Well, if I'm not mistaken this command is supposed to "deploy" a package from a workspace in the sense of making this independent. In other words, it should have all dependencies present in the local `node_modules` in a way we can execute that without depending on the workspace. 
If we workaround the above permission problem by changing the root user with `USER` and setting the directories permissions with `chown` on Dockerfile, it doesn't include the necessary dependencies like `@nestjs/core` on the `node_modules`. So when we run `pnpm deploy` the exported `node_modules` doesn't have the expected deps. I found a similar issue on pnpm repository https://github.com/pnpm/pnpm/issues/5415

## Next steps
I know you might be thinking that we have a bunch of duplicated code and I also know it's possible to use a Dockerfile on the GitHub action. But I wanna keep stuff simple for now and not overengineer it. So for the next steps, if we need to keep this workaround I want to dedupe a little bit.

- [ ] Try to dedupe by reusing Dockerfile on GH Actions
- [ ] Run e2e tests for backend only when backend package has changed
- [ ] Create expo preview only when mobile package has changed


apps/backend/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM node:14 as base
```
abarbosa-godaddy, line 1: Why are you doing a multistage build?

docker-compose.yml
```diff
@@ -1,7 +1,18 @@
...
+  backend:
+    build:
+      context: .
+      dockerfile: apps/backend/Dockerfile
+    depends_on:
+      - mongodb
+    ports:
+      - 3000:3000
+    environment:
+      - MONGODB_URI=mongodb://mongo:mongo@animavita_mongo/?authMechanism=DEFAULT
```
abarbosa-godaddy, line 13: NIT: you could use env files for that

https://towardsdatascience.com/a-complete-guide-to-using-environment-variables-and-files-with-docker-and-compose-4549c21dc6af?gi=3540cd2edfc1

.npmrc-default-node-linker
```diff
@@ -0,0 +1 @@
+auto-install-peers=true
```
abarbosa-godaddy: EOF

apps/backend/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm
+
+WORKDIR /app
+
+# pnpm fetch does require only lockfile
+COPY pnpm-lock.yaml .
+RUN pnpm fetch
+
+ADD . .
```
abarbosa-godaddy: NIT: To avoid unexpected results you could use COPY here.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#:~:text=ADD%20or%20COPY&text=COPY%20only%20supports%20the%20basic,rootfs.tar.xz%20%2F%20.

.dockerignore
```diff
@@ -0,0 +1,14 @@
...
+.git
+.gitignore
+npm-debug.log
+.dockerignore
+Dockerfile
+
+# Directories
+node_modules
+test
+dist
```
abarbosa-godaddy: EOF

apps/backend/Dockerfile
belgamo: Because we need first to prepare the workspace stuff as it is a monorepo.

belgamo: I left some references in the PR description in case you wanna dive deeper into it.

Issue #128, opened 2022-11-04
Create user & auth module


belgamo:
### Task
Let's add a module for our users. It's very self-explainable so here I'll just mention which fields it should contain:

- name
- email
- location (coordinates)
- photoUri

### [Authorization](https://docs.nestjs.com/security/authentication)
Let's add an authorization step where users can log in with their e-mail and password.

JVBorges:
Can I take this one?

belgamo:
> Can I take this one?

Please!

JVBorges:
which strategy will be using for authentication?

PR #135, opened 2022-11-13
Create user and auth module

JVBorges:
# Summary

Address https://github.com/animavita/animavita/issues/128

This PR consists of adding the resources for registration and login of users. JWT with refresh tokens was the strategy adopted for the authorization of users.

## Tasks

- [x] Add unit test
- [x] Add integration test


apps/backend/src/user/entities/mongodb/user.schema.ts
```diff
@@ -0,0 +1,17 @@
...
+export const UserSchema = new mongoose.Schema<UserType>(
+  {
+    name: String,
+    email: String,
+    password: String,
+    location: {
+      longitude: Number,
+      latitude: Number,
+    },
+    photoUri: String,
```
belgamo: Should be optional

apps/backend/src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+          useValue: userServiceMock,
+        },
+      ],
+    }).compile();
+
+    service = module.get<AuthService>(AuthService);
+    jwtService = module.get<JwtService>(JwtService);
+    configService = module.get<ConfigService>(ConfigService);
+    userService = module.get<UserService>(UserService);
+  });
```
belgamo: Not a big deal, that's fairly well. However, I am a big fan of taking a more functional approach and creating a setup function and get rid of your `beforeEach`, something similar to this:

Outside of describe:

```js
const setup = () => {
 const module = await Test.createTestingModule({ ... })

 service = module.get<AuthService>(AuthService);
 jwtService = module.get<JwtService>(JwtService);

 return {
  service,
  jwtService,
 }
}
```

Then, on each test:
```js
it('should be defined', () => {
 const { service } = setup();

 expect(service).toBeDefined();
});
```

This way you ensure you have a totally new variable (any other test did not use that). I know your `beforeEach` is performing this job but I really think that using the `setup` approach makes your test more explicit for other developers.

WDYT?

apps/backend/src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    });
+
+    it('should generate new pair of tokens', async () => {
+      jest.spyOn(userService, 'findById').mockResolvedValueOnce({
+        ...user,
+        refreshToken: await argon.hash(tokens.refreshToken),
+      });
+
+      // I need to wait a second because if I don't the tokens will be the same
+      await new Promise((r) => setTimeout(r, 1000));
```
belgamo: Try to mock the module responsible for making you await as we talked about on discord.

apps/backend/src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        service.refreshTokens(user.id, 'different_token'),
+      ).rejects.toThrowError(ForbiddenException);
+    });
+
+    it('should throw if given token its not the current', async () => {
+      jest.spyOn(userService, 'findByEmail').mockResolvedValueOnce(user);
+      jest.spyOn(userService, 'update').mockResolvedValueOnce(null);
+
+      // I need to wait a second because if I don't the tokens will be the same
+      await new Promise((r) => setTimeout(r, 1000));
```
belgamo: Same as above.

apps/backend/src/user/user.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  async create(user: UserType) {
+    const foundUser = await this.userRepository.findByEmail(user.email);
+
+    if (foundUser)
+      throw new UnprocessableEntityException('Email already taken');
+
+    return this.userRepository.create(user);
+  }
+
+  findById(userId: string) {
```
belgamo, line 27: Don't we need to implement a conditional for the case where the user is not found and then throw an error, as you did for `findById`?

apps/backend/test/e2e/auth.e2e-spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      .signUp(userMock)
+      .then(() =>
+        service.signIn({
+          email: userMock.email,
+          password: userMock.password,
+        }),
+      );
+
+    // I need to wait a second because if I don't the tokens will be the same
+    await new Promise((r) => setTimeout(r, 1000));
```
belgamo: Same as the other timeouts you have

shared/models/user.ts
```diff
@@ -0,0 +1,13 @@
+export type Coordinates = { latitude: number; longitude: number };
+
+export type UserType = {
+  id?: string;
+  name: string;
+  email: string;
+  password: string;
+  location: Coordinates;
+  photoUri: string;
```
belgamo: Let's make it optional since we don't want to require users to upload a picture while signing up.

shared/validation-schemas/src/auth.ts
```diff
@@ -0,0 +1,13 @@
...
+  .fork(["id"], (schema) => schema.forbidden())
+  .fork(["name", "email", "password", "location", "photoUri"], (schema) =>
+    schema.required()
+  );
+
+export const signInValidationSchema = userValidationSchema
+  .fork(["id"], (schema) => schema.forbidden())
+  .fork(["email", "password"], (schema) =>
+    schema.required()
+  );
```
belgamo: EOF

apps/backend/test/e2e/auth.e2e-spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        AuthModule,
+        ConfigModule.forRoot({ isGlobal: true }),
+      ],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    service = moduleFixture.get<AuthService>(AuthService);
+
+    await app.init();
+  });
```
belgamo: Take a look at my comment about `setup` approach.

apps/backend/src/user/user.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    return this.userRepository.findById(userId);
+  }
+
+  findByEmail(email: string) {
+    return this.userRepository.findByEmail(email);
```
belgamo, line 32: Same as above

apps/backend/src/user/user.service.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+        {
+          provide: 'MONGODB',
+          useValue: repositoryMock,
+        },
+      ],
+    }).compile();
+
+    service = module.get<UserService>(UserService);
+    repository = module.get<UserRepository>('MONGODB');
+  });
```
belgamo: Take a look at my comment about `setup` approach.

belgamo (changes requested):
Thanks for contributing. There are a few things we need to tweak but overall looks good!

apps/backend/src/user/user.service.ts
JVBorges: With these simple queries function, I prefer to leave the validation responsibility to the calling function. I found it easier than using a try-catch.

apps/backend/src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        password: userMock.password,
+      });
+
+      expect(tokens).toEqual(
+        expect.objectContaining({
+          accessToken: expect.any(String),
+          refreshToken: expect.any(String),
+        }),
+      );
+      expect(spiedFunction).toBeCalled();
```
belgamo: `toBeCalledTimes` to make sure it was called only once.

apps/backend/src/user/user.service.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      const { service, repository } = await setup();
+      jest.spyOn(repository, 'findByEmail').mockResolvedValueOnce(userMock);
+      await expect(service.create(userMock)).rejects.toThrowError(
+        UnprocessableEntityException,
+      );
+    });
+  });
+
+  describe('findById', () => {
+    it('should find a user by his id', async () => {
```
belgamo:
```suggestion
    it('should find a user by their id', async () => {
```

apps/backend/src/user/user.service.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      const { service, repository } = await setup();
+
+      jest.spyOn(repository, 'findById').mockResolvedValueOnce(userMock);
+      await expect(service.findById(userMock.id)).resolves.toEqual(userMock);
+    });
+  });
+
+  describe('findByEmail', () => {
+    it('should find a user by his email', async () => {
```
belgamo:
```suggestion
    it('should find a user by their email', async () => {
```

apps/backend/src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    id: 'someId',
+    refreshToken: 'oldToken',
+    ...userMock,
+  };
+
+  const userServiceMock = {
+    create: jest.fn().mockResolvedValue(persistedUser),
+    findById: jest.fn().mockResolvedValue(persistedUser),
+    findByEmail: jest.fn().mockResolvedValue(persistedUser),
+    update: jest.fn().mockResolvedValue(persistedUser),
```
belgamo, line 28: Do you really need `jest.fn()` here? What about just `create: Promise.resolve(persistedUser)`

apps/backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,100 @@
...
+  private async generateTokens(userId: string, email: string) {
+    const [accessToken, refreshToken] = await Promise.all([
+      this.jwtService.signAsync(
+        {
+          sub: userId,
+          email,
+        },
+        {
+          secret: this.configService.get<string>('JWT_ACCESS_SECRET'),
+          expiresIn: '1h',
```
belgamo: Could you use a constant for that?

apps/backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        },
+      ),
+      this.jwtService.signAsync(
+        {
+          sub: userId,
+          email,
+        },
+        {
+          secret: this.configService.get<string>('JWT_REFRESH_SECRET'),
+          expiresIn: '7d',
```
belgamo: Same as above

apps/backend/src/user/repositories/mongodb/user.schema.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const UserSchema = new mongoose.Schema<UserType>(
+  {
+    name: String,
+    email: String,
+    password: String,
+    location: {
+      longitude: Number,
+      latitude: Number,
+    },
```
belgamo: What about [geospatial-queries](https://www.mongodb.com/docs/manual/geospatial-queries/)?

apps/backend/test/e2e/auth.e2e-spec.ts
```diff
@@ -0,0 +1,131 @@
...
+      .post('/api/v1/auth/signUp')
+      .send(userMock)
+      .expect(201);
+
+    expect(body).toStrictEqual({
+      message: 'User created successfully',
+    });
+
+    await app.close();
+  });
```
belgamo: Not a big deal but you nest describe blocks to make the expected output explicit:

```js
describe('/POST auth/signUp', () => {
 it('returns a successful message', () => { ... })
})
```

apps/backend/src/auth/auth.service.spec.ts
JVBorges: Not really I think, but why choosing this way instead?

belgamo: Just because it's simpler and in your test you don't have any `toBeCalled` check hehe

apps/backend/.env.example
```diff
@@ -1 +1,6 @@
-MONGODB_URI=mongodb://mongo:mongo@localhost:27017/animavita?authSource=admin
+MONGODB_URI=mongodb://mongo:mongo@localhost:27017/?authMechanism=DEFAULT
+
+JWT_ACCESS_SECRET=
+JWT_ACCESS_TOKEN_EXPIRATION=24h
+JWT_REFRESH_SECRET=
+JWT_REFRESH_TOKEN_EXPIRATION=14d
```
belgamo, line 6: EOF

belgamo (approved):
Please rebase your branch with `v2`

belgamo (dismissed):

belgamo (dismissed):

belgamo (approved):

belgamo:
Thanks a lot @JVBorges, we did a big progress with this PR 👊

PR #150, opened 2023-01-16
Manage adoption form state with react-hook-form

belgamo:
closes https://github.com/animavita/animavita/issues/147

https://user-images.githubusercontent.com/19699724/212688344-8dc97b2b-3a8c-4468-8f74-b8a90a78b7fe.mp4

### Integrating react-hook-form with NativeBase
I created a new directory `/components/react-hook-form/native-base` to place form components related to RHF. So basically whenever we want to control a `Input` using RHF, we just need to import `RHFNativeBaseInput` which is the integration between the two libraries.

### Validation flow
RHF provides certain [modes](https://react-hook-form.com/api/useform) to trigger validation. We're using the `onBlur` validation mode (although it doesn't perform any action) and also triggering the validation manually when the user clicks on "Next step". There is a custom hook to encapsulate all this validation logic which means that we call the `validateField` returned from `useFormValidation` and we only proceed to the next step if the current one is valid. If it's not valid, we show a toast message informing what's wrong to the user. 

#### Validation messages
For now, we're using the default error messages from joi. I know it's not so friendly right now, so we do want to come up with better messages and also translate them into Portuguese. Created a new task for that #151

#### ESLint & Prettier
Was having conflicts with both libraries, I fixed that.

#### TextEncoder issue
After import the joi schema from `@animavita/validation-schemas`, I started facing an issue with `TextEncoder`. Solved by installing `fast-text-encoding` package.  See https://github.com/hapijs/joi/issues/2141

#### Expo preview
Had to add an extra step to `setup-morepo` GH action to build shared dependencies since expo was finding the index entries for them.


.github/actions/setup-monorepo/action.yml
```diff
@@ -46,3 +46,7 @@ runs:
     - name: 📦 Install dependencies
       run: pnpm install
       shell: bash
+
+    - name: Build shared dependencies
+      run: pnpm --filter "./shared/**" run build
+      shell: bash
```
brunohkbx: Missing EOF

apps/mobile/src/components/react-hook-form/native-base/rhf-list-selector.tsx
```diff
@@ -0,0 +1,37 @@
...
+  label: string;
+  value: string;
+};
+
+type RHFNativeBaseListSelectorProps = {
+  name: string;
+  options: Option[];
+};
+
+function RHFNativeBaseListSelector({ name, options }: RHFNativeBaseListSelectorProps) {
```
brunohkbx, line 15: Does Animavita is following/enforcing a pattern when it comes to arrow functions or normal functions? If not, can we enforce it with eslint?

apps/mobile/src/components/react-hook-form/native-base/rhf-native-input.tsx
```diff
@@ -0,0 +1,20 @@
+import { Input } from 'native-base';
+import { IInputProps } from 'native-base/lib/typescript/components/primitives/Input/types';
```
brunohkbx: Looks like native-base export this type
https://github.com/GeekyAnts/NativeBase/blob/f85d0a64e536d6205a170c91dd0d1e5441a5bdd0/src/index.tsx

apps/mobile/src/components/react-hook-form/native-base/rhf-native-input.tsx
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react';
+import { useController } from 'react-hook-form';
+
+type RHFNativeBaseInputProps = {
+  input: IInputProps;
+  control: any;
+  name: any;
```
brunohkbx: sounds like it could be a string

apps/mobile/src/components/register-adoption-form/adoption-form.component.tsx
```diff
@@ -8,25 +11,41 @@ import StepperIndicator from './compose/stepper-indicator';
 
 export default function RegisterAdoptionForm() {
   const { activeStep, handleBack, handleNext, isLastStep, isFirstStep } = useMultiStepNavigation();
+  const adoptionForm = useForm({
+    defaultValues: {
+      name: '',
+      gender: '',
+      breed: '',
+      type: '',
+      age: 1,
```
brunohkbx, line 20: Why 1? I'd move this to a const at the top like `MIN_AGE` to give more visibility and context

shared/validation-schemas/src/adoption.ts
```diff
@@ -1,3 +1,4 @@
+import "fast-text-encoding";
```
brunohkbx, line 2: just out of curiosity, why do we need this?

belgamo: left a link in the PR description https://github.com/hapijs/joi/issues/2141

apps/mobile/src/components/react-hook-form/native-base/rhf-list-selector.tsx
belgamo: no we're not. yes, I'll set lint rules to prevent that.

apps/mobile/src/components/register-adoption-form/adoption-form.component.tsx
belgamo: that's a very good question. actually I don't even know if a slider component makes sense to set the age cause we should consider newborn pets, which means that they'd have less than 1 year. I think we can go back to this in a future PR to come up with a better solution/component to handle that.

shared/validation-schemas/src/adoption.ts
brunohkbx: WDYT about adding a small comment at the top for future reference?

apps/mobile/src/components/register-adoption-form/adoption-form.component.tsx
belgamo: https://github.com/animavita/animavita/issues/152

shared/validation-schemas/src/adoption.ts
belgamo: yep, sounds good.

apps/mobile/src/components/react-hook-form/native-base/rhf-list-selector.tsx
belgamo: Added the rule with `warn` for now, otherwise, I'd have to change a thousand of files. I'll convert the `function` declarations into arrow format in the next PR.

PR #165, opened 2023-03-05
Build authentication frontend logic

belgamo:
### Overview
This is supposed to create the logic to show the sign-in or internal screens such as home, adoption register, etc. In a nutshell, I added a new `AuthProvider` that is responsible to handle the authentication flow and to know whether render the sign-in or the internal routes, we basically have a condition in the `MainNavigator` component watching the `auth.status` state.

### Storage approach
I decided to use `expo-secure-store` to store the user's JWT token since it seems to be the most popular approach being used.

### On web
For now, since we won't have a lot of users coming from the web, I decided not to store the user's token on any storage system because doing so, we're susceptible to XSS attacks and our backend doesn't support an HttpOnly cookie, which would be ideally better. Therefore, for now, we're gonna have the token in memory, and users that want to use Animavita on the web will have to log in every time they refresh the page.

### Platform-specific code
As you may already noticed, we're taking different auth approaches depending on where the user is using the app. For this reason, I tried to avoid conditionals and isolated the code for native and web in two different files: `use-auth-actions.native.ts` and `use-auth-actions.native.ts`. This way we can easily change the code for one platform without impacting another.

### DIFF
--- a/apps/mobile/src/screens/signin/signin.screen.tsx
+++ b/apps/mobile/src/screens/signin/signin.screen.tsx
@@ -1,0 +1,27 @@
+import { Button, View } from 'native-base';
+
+import AppStatusBar from '../../shared/components/status-bar/status-bar.component';
+import { useAuth } from '../../shared/hooks/use-auth-provider';
+
+const SignInScreen = () => {
+  const auth = useAuth();
+
+  return (
+    <View flex="1" alignItems="center" justifyContent="center">
+      <AppStatusBar />
+      <Button
+        variant="outline"
+        onPress={() => {
+          auth.signIn({
+            accessToken: '1234',
+            refreshToken: '4321',
+          });
+        }}
+      >
+        Entrar
+      </Button>
+    </View>
+  );
+};
+
+export default SignInScreen;
--- a/apps/mobile/src/screens/signin/signin.screen.tsx
+++ b/apps/mobile/src/screens/signin/signin.screen.tsx
@@ -2,9 +2,11 @@ import { Button, View } from 'native-base';
 
 import AppStatusBar from '../../shared/components/status-bar/status-bar.component';
 import { useAuth } from '../../shared/hooks/use-auth-provider';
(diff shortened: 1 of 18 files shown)


gquaresma-godaddy:
LGTM!

gquaresma-godaddy (dismissed):

joaoGabriel55:
LGTM!

apps/mobile/src/screens/signin/signin.screen.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return (
+    <View flex="1" alignItems="center" justifyContent="center">
+      <AppStatusBar />
+      <Button
+        variant="outline"
+        onPress={() => {
+          auth.signIn({
+            accessToken: '1234',
+            refreshToken: '4321',
```
brunohkbx, line 19: Can we have variables for these values at the top? just to give more context with a meaningful name, it's not clear to me why these values are used here.

apps/mobile/src/screens/signin/signin.screen.tsx
```diff
@@ -0,0 +1,27 @@
...
+          });
+        }}
+      >
+        Entrar
```
brunohkbx: Shouldn't this use i18n?

apps/mobile/src/screens/splash/splash.screen.tsx
```diff
@@ -0,0 +1,11 @@
+import { Text, View } from 'native-base';
+
+const SplashScreen = () => {
+  return (
+    <View flex="1" alignItems="center" justifyContent="center" bg="danger.800">
+      <Text>Carregando...</Text>
```
brunohkbx: Same

apps/mobile/src/screens/signin/signin.screen.tsx
belgamo: This is just mock data to simulate the login for now. Later, we will replace it with a call to the backend endpoint.

PR #171, opened 2023-04-13
Hot-reload for shared packages

belgamo:
### Summary
While developing, we manually refreshed the app when some change was made to the shared packages. In order to improve the DX a little bit, I set up [NX](https://nx.dev/) to watch changes to the shared packages and then rebuild it whenever a change is made. I also defined a `postinstall` instruction so our first-time contributors don't need to build the shared packages manually.
NX might seem too much for now for such a simple task, but it will help us with other things in the near future.

### Backend
Unfortunately, for the backend packages, we still need to rebuild the docker image if a change is made to the shared packages. I haven't found a way to auto-update it since we don't map the `node_modules` with the docker-compose volumes.
We also can't use the alpine image anymore because NX depends on some Linux libraries. It's gonna turn our build heavier.


joaoGabriel55 (approved):
LGTM

PR #173, opened 2023-05-01
Add ability to find nearest adoptions

### LINKED ISSUE #172: Add ability to find near adoptions
Since it doesn't make sense to suggest adoptions from the US to a Brazilian user, we need a way to look for the nearest adoptions and filter out the rest.

belgamo:
closes #172 

This PR adds one of the core functionalities of this app which is to be able to find pets within a given radius.

Screen Shot 2023-05-01 at 12 12 13

In the above example, the red point is a user looking up for pets to adopt. We should give this user adoption suggestions within a radius of 1km which means that pets registered by users living on 3rd Ave or any other place outside of the circle wouldn't be returned.

### Summary
- Saved the user reference in the adoption document
- Added `/nearMe?` endpoint
- Decoupled mongo from the adoption module (repository pattern)
- Renamed the ~~adoptions~~ module to `adoption`

### Relationship between users and adoptions
Now whenever the user creates an adoption, we store the user id in the adoption document. This way we know what adoptions belong to a given user. Also, we replicate the `location` object from the user and store it in the adoption so we don't need to retrieve the user info to filter out adoptions outside of a radius.

### Geospatial queries
MongoDB has a built-in feature to perform this kind of lookup. In our case, we used the [`$center`](https://www.mongodb.com/docs/manual/reference/operator/query/center/) to find the nearest pets within the radius the user wants. So later when we start working on the frontend part, we can use a [range component](https://docs.nativebase.io/slider) in kilometers.
Also, we don't want to return adoptions owned by the user who is searching. This is an important business rule because it doesn't make sense at all. That's why we use `.ne(currentUser.id)` meaning to find adoptions not belonging to the logged user.

### Testing strategy
We already use [mongodb-memory-server](https://github.com/nodkz/mongodb-memory-server) to test stuff. So we basically wrote a new e2e test case with two users living in the same city. One user wants to adopt and another one can't take care of their pet anymore.


joaoGabriel55 (approved):
LGTM

abarbosa-godaddy:
WOW!

abarbosa-godaddy (approved):

apps/backend/src/adoption/adoption.interface.ts
```diff
@@ -0,0 +1,12 @@
+import { AdoptionType } from '@animavita/models';
+import { Location } from '../user/repositories/mongodb/user.interface';
+
+export type AdoptionDto = AdoptionType;
+
+export type AdoptionResponse = AdoptionType & {
+  user: {
+    id: string;
```
brunohkbx, line 8: After skimming mongodb documentation, I found this ObjectId type, does it make sense to use it here?
https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/#insert-operations-and-the-_id-field

apps/backend/src/adoption/repositories/adoption-repository.interface.ts
```diff
@@ -0,0 +1,23 @@
+import { AdoptionType, UserType } from '@animavita/models';
+import { AdoptionResponse } from '../adoption.interface';
+import { Location } from '../../user/repositories/mongodb/user.interface';
+
+export type IAdoption = AdoptionType & {
+  user: string;
```
brunohkbx, line 6: What does `user` means in this case? Is it an ID? Or the user name? 
Can we rename it to something else?

apps/backend/test/e2e/adoptions.e2e-spec.ts
```diff
@@ -122,6 +161,69 @@ describe('AdoptionsController (e2e)', () => {
...
+
+    it('returns all adoptions within a radius of 1km', async () => {
+      return request(app.getHttpServer())
+        .get('/api/v1/adoptions/nearMe?radius=1')
+        .auth(authToken, { type: 'bearer' })
+        .expect(200)
+        .expect((res) => {
+          return expect(res.body).toEqual([
+            expect.objectContaining({
+              ...pet2Mock,
```
brunohkbx, line 182: Why only the pet2Mock is being returned? Can we explicitly pass the coordinates in the setup block to make it clear that only pet2 is within this radius?
BTW this library is pretty cool to create factories:
https://github.com/thoughtbot/fishery

apps/backend/src/adoption/repositories/mongodb/adoption-mongo.repository.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const RADIUS_OF_EARTH = 63781; // km
+
+export class AdoptionMongoDBRepository implements AdoptionRepository {
+  constructor(
+    @InjectModel('Adoption')
+    private readonly adoptionModel: Model<AdoptionDocument>,
+  ) {}
+
+  async create(adoption: IAdoption) {
```
brunohkbx, line 19: What happens if the creating process fails due to invalidate data? How do you handle that back to the user?

apps/backend/src/adoption/repositories/mongodb/adoption-mongo.repository.ts
```diff
@@ -0,0 +1,74 @@
...
+    const newAdoption = new this.adoptionModel(adoption);
+
+    const document = await newAdoption
+      .save()
+      .then((adoption) => adoption.populate('user', 'id name'));
+
+    return AdoptionMap.toType(document);
+  }
+
+  async update(adoption: Partial<IAdoption>) {
```
brunohkbx, line 29: Same question

apps/backend/src/adoption/adoption.interface.ts
belgamo: We have a few types of interfaces; in this case, it's just the interface used in the service response. So basically it's not related to Mongo itself but to the JSON schema to be returned. Now if you look at [adoption-schema](https://github.com/animavita/animavita/blob/v2/apps/backend/src/adoption/repositories/mongodb/adoption-mongo.schema.ts) we do have a reference to the user using object id type.

apps/backend/src/adoption/repositories/adoption-repository.interface.ts
belgamo: Yeah, this is supposed to be the `userId`, good catch!

apps/backend/test/e2e/adoptions.e2e-spec.ts
belgamo: Screen Shot 2023-05-07 at 10 46 39

Because that's the pet we link to the user on the setup block. Usually, we wouldn't have to create adoptions to test the GET endpoint. However, I haven't found a standard way to start the in-memory mongo with a db dump or something like that. Then because of the `authToken`, we know which user to get adoptions from according to the coordinates.

Screen Shot 2023-05-07 at 10 51 41

Basically, during the creation step, we get the user coordinates and store them in the adoption document. That's why only pet2 is returned.

Please, lemme know if you have any suggestions to make it clear

apps/backend/src/adoption/repositories/mongodb/adoption-mongo.repository.ts
belgamo: Screen Shot 2023-05-07 at 10 59 28

Nest has a default error validator. In case of missing data, the repository layer will throw and the service layer will handle that as a 400 error.

Issue #129, opened 2022-11-04
Add login & onboarding screens


belgamo:
depends on #128 

#### Login screen
Let's add a [new screen](https://reactnavigation.org/docs/stack-navigator/) so users can type their e-mail and password to log in.

#### Location
The main goal of this app is to match people who want to adopt a pet with people who want to put a pet for adoption. Of course, we don't want to have matches between a person from the North with a person from the South of the country. In order to solve this, let's [capture the user's location](https://docs.expo.dev/versions/latest/sdk/location/) when they first log into our app and store it on the backend so we can use it to filter the adoptions feed.

### Screenshots/examples
| | | | |
:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:|
![image](https://user-images.githubusercontent.com/19699724/199883729-598cdc73-1ac7-4eb8-9023-ab55c18e1a03.png)| ![image](https://user-images.githubusercontent.com/19699724/199883913-01c62150-5f5d-4e01-ba04-27a8b92aa57d.png) | ![image](https://user-images.githubusercontent.com/19699724/199884070-1e46a95d-ac92-4064-9e85-5410056c556c.png) | ![image](https://user-images.githubusercontent.com/19699724/199884186-1c9d1417-e995-4f7f-9449-ac6b9f95a7d3.png) |

joaoGabriel55:
I can take this

Isaac-alencar:
Hello, I'd like to work on it with @bruno.viveiros as co-author if it's possible

belgamo:
Hey, sure, go ahead! I updated the issue description and removed the secure token storage requirement since it's already done https://github.com/animavita/animavita/pull/165

BrunoViveiros:
Thanks

belgamo:
Hey guys, any update on this? Can I reassign it?

Isaac-alencar:
> Hey guys, any update on this? Can I reassign it?

Hey, sorry for delaying this task. Can I still work on it? I can deliver by the end of the next week. sounds good?

belgamo:
It's done! Thank you guys!

PR #181, opened 2023-08-08
Feat/Add onboarding and signin screen

Isaac-alencar:
This PR aim to close the https://github.com/animavita/animavita/issues/129 issue

###  What I did?
 - Added a new screen to capture the user current location;
 - Updated the visual of Sign In screen by adding fields to type email and password;


apps/mobile/src/hooks/use-geolocation/index.ts
```diff
@@ -0,0 +1,67 @@
+import * as Location from 'expo-location';
+import { useState } from 'react';
+import { Alert, Linking } from 'react-native';
+
+export const useGeolocation = () => {
```
belgamo: Missing tests... What about having a unit test to check the output state of this hook? In other words, a unit tests ensuring the values of `address` and `isLoading`

apps/mobile/src/hooks/use-geolocation/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  const [address, setAddress] = useState<Location.LocationGeocodedAddress[]>();
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const errorAlert = (msg: string, onPress: () => void, text?: string) =>
```
belgamo: This hook shouldn't be responsible for emitting alerts. This is a UI/view concern.

apps/mobile/src/hooks/use-geolocation/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      if (!hasServicesEnabled) {
+        errorAlert('Por favor, verifique se o seu GPS está habilitado!', () => {
+          Location.enableNetworkProviderAsync();
+        });
+
+        return;
+      }
+
+      if (!granted && !canAskAgain) {
```
belgamo: This logic should be placed in the component

apps/mobile/src/hooks/use-geolocation/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        errorAlert(
+          'A sua localização é realmente necessária para uma melhor experiência no app.',
```
belgamo: This should be in the translation file

apps/mobile/src/hooks/use-geolocation/index.ts
```diff
@@ -0,0 +1,67 @@
...
+          () => {
+            Linking.openSettings();
+          },
+          'Go to settings'
```
belgamo: Same

apps/mobile/src/hooks/use-geolocation/use-geolocation.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Alert } from 'react-native';
+
+import { useGeolocation } from '.';
+import { getLocationModuleMock } from './fixtures/mocks/location';
+import GetLocation from '../../screens/get-location/get-location.screen';
+import { renderWithProviders } from '../../test/test-utils';
+
+jest.mock('expo-location');
+
+describe('useGeolocation', () => {
```
belgamo: This is actually testing the `GetLocationScreen`, not the `useGeoLocation` hook

apps/mobile/src/i18n/locales/pt-BR.json
```diff
@@ -68,5 +71,14 @@
...
     "LOGOUT": "Sair"
+  },
+  "SHARE_LOCATION": {
+    "GREETINGS": "Olá, {{ name }} precisamos da sua",
+    "WHERE_USER_ARE_MSG": "Você está em",
+    "CONFIRM_BUTTON": "Confirmar minha localização",
+    "GET_LOCATION": "Compartilhar minha localização",
+    "GET_LOCATION_NEW_ATTEMPT": "Buscar novamente",
+    "LOCATION": "localização",
+    "IMAGE_ALT_TEXT": "A boy waving with his dog besides him"
```
belgamo: Please, translate to portuguese

apps/mobile/src/navigation/main-navigator.tsx
```diff
@@ -31,6 +32,7 @@ const MainNavigator = () => {
             <Stack.Screen name={Routes.SignIn} component={SignInScreen} />
           </>
         )}
+        <Stack.Screen name={Routes.GetLocation} component={GetLocation} />
```
belgamo: Let's not render the GetLocation screen since it should be rendered only on the user onboarding (signup). I'll address another issue for that

apps/mobile/src/screens/signin/compose/signin-form.tsx
```diff
@@ -0,0 +1,53 @@
...
+        marginTop={10}
+        width="full"
+        onPress={() => {
+          auth.signIn({
+            accessToken: '1234',
+            refreshToken: '4321',
+            name: 'John Doe',
+          });
+
+          navigate(Routes.GetLocation as never);
```
belgamo: You shouldn't redirect the user to the GetLocation after logging in, but after filling out the signup form (will have another issue for that purpose).

apps/mobile/src/hooks/use-geolocation/index.ts
```diff
@@ -0,0 +1,67 @@
+import * as Location from 'expo-location';
+import { useState } from 'react';
+import { Alert, Linking } from 'react-native';
+
+export const useGeolocation = () => {
+  const [address, setAddress] = useState<Location.LocationGeocodedAddress[]>();
```
belgamo: I don't recommend you use this type here because this way you're coupling expo-location to the code. It's better to define your own interface so that we can easily switch expo-location in the future in case we need it.

apps/mobile/src/hooks/use-geolocation/index.ts
belgamo: `useGeolocation` doesn't seem to be a good name since it's too generic.

apps/mobile/src/screens/get-location/compose/action-buttons-group.tsx
```diff
@@ -0,0 +1,42 @@
+import { Button, Spinner, View } from 'native-base';
+
+import useLocale from '../../../hooks/use-locale';
+
+type ButtonGruopProps = {
```
belgamo: typo

belgamo (changes requested):

belgamo:
You added the `GetLocation` screen but not the `Signup`. I'll address a new issue; for now, we don't use the location components. Thank you for this awesome PR 👊

PR #182, opened 2023-09-07
Refactor shared types 

belgamo:
I noticed that by trying to reuse a single type definition for everything (dto, db entities, responses) we were falling into a trap. It's better to have separate types (although it feels like we're duplicating) because they represent and mean different things.

Remember that our final goal in having shared interfaces is to be able to guarantee the contract between the front and backend.


apps/backend/package.json
```diff
@@ -14,14 +14,14 @@
...
-    "test": "jest",
-    "test:watch": "jest --watch",
-    "test:cov": "jest --coverage",
-    "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json"
+    "test": "TZ=UTC jest",
+    "test:watch": "TZ=UTC jest --watch",
+    "test:cov": "TZ=UTC jest --coverage",
+    "test:debug": "TZ=UTC node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
+    "test:e2e": "TZ=UTC jest --config ./test/jest-e2e.json"
```
brunohkbx: I wonder if wouldn't be better to have this env declared in the jest configuration section at this file

apps/backend/src/user/repositories/user-repository.interface.ts
```diff
@@ -1,9 +1,36 @@
...
-import { IUser } from './mongodb/user.interface';
+import { Exclude } from 'class-transformer';
+
+export class UserEntity {
+  id: string;
+  name: string;
+  email: string;
+
+  @Exclude()
+  password: string;
```
brunohkbx, line 9: Is there any alternative way of doing the same without needing a new package? I wonder if Omit works here. Found this example: https://www.prisma.io/docs/concepts/components/prisma-client/excluding-fields#excluding-the-password-field

apps/backend/src/adoption/repositories/adoption-repository.interface.ts
```diff
@@ -1,23 +1,50 @@
...
-export type IAdoption = AdoptionType & {
-  user: string;
-  location: Location;
+export class AdoptionEntity {
+  id: string;
+  name: string;
+  gender: AnimalGendersType;
+  breed: string;
+  type: AnimalType;
+  age: AnimalAgesType;
```
brunohkbx, line 15: Out of the scope of this PR but I feel like `age` sounds confusing here as in implying to be a numeric value. I wonder if there's a better name to represent the values of this enum

apps/backend/src/adoption/repositories/mongodb/adoption-mongo.repository.ts
```diff
@@ -16,39 +19,46 @@ export class AdoptionMongoDBRepository implements AdoptionRepository {
...
-    const newAdoption = new this.adoptionModel(adoption);
+  async create(adoption: AdoptionEntity) {
+    const newAdoption = new this.adoptionModel(AdoptionMap.toSchema(adoption));
 
     const document = await newAdoption
       .save()
-      .then((adoption) => adoption.populate('user', 'id name'));
+      .then((adoption) =>
+        adoption.populate<PopulatedAdoptionDocument>('user', 'id name'),
+      );
```
brunohkbx: Does it work without this `then` call? just noticed that the function below doesn't have it

```suggestion
      .populate<PopulatedAdoptionDocument>('user', 'id name')
```

apps/backend/test/e2e/adoptions.e2e-spec.ts
```diff
@@ -21,6 +21,8 @@ import { AuthService } from '../../src/auth/auth.service';
 import { ConfigModule } from '@nestjs/config';
 import { getCoordinatesFromUser } from '../../src/user/user.helpers';
 
+Date.now = () => 1000;
```
brunohkbx: Is this change intentional?

apps/backend/test/mocks/user.ts
```diff
@@ -1,6 +1,20 @@
...
+  password: 'quesenharuimgrosbilda',
+  location: {
+    latitude: 10.002322,
+    longitude: -40.343569,
+  },
+  photoUri: 'http://www.google.com/',
+  id: '123',
+  createdAt: '',
+  updatedAt: '',
+};
```
brunohkbx, line 15: Also out of the scope of this PR but would be nice to set up [Faker](https://fakerjs.dev/) for that

belgamo: Thanks, addressing an issue for that https://github.com/animavita/animavita/issues/185

apps/backend/src/adoption/repositories/mongodb/adoption-mongo.repository.ts
belgamo: Yep, it does. Good catch!

apps/backend/src/adoption/repositories/adoption-repository.interface.ts
belgamo: It used to be a number. However, we decided to come up with a generic way of representing the age of the pet since it might be a dog, a cat, or any other animal. Right now I can't think of a better name for that, but it makes sense. Addressing an issue for that https://github.com/animavita/animavita/issues/186

apps/backend/package.json
belgamo: For sure, thanks!

apps/backend/src/user/repositories/user-repository.interface.ts
belgamo: The advantage of using this approach is that Nest automatically excludes the desired properties through interceptors https://docs.nestjs.com/techniques/serialization#exclude-properties

apps/backend/test/e2e/adoptions.e2e-spec.ts
belgamo: It was a quick test and I forgot to remove. Thanks

PR #187, opened 2023-09-13
add fishery and faker, rewrite mock objects to use the same

Vedu1996:


apps/backend/test/factories/adoption.ts
```diff
@@ -0,0 +1,20 @@
+import { Factory } from 'fishery';
+import { faker } from '@faker-js/faker';
```
belgamo, line 2: Please include these dependencies in the app's package.json.

belgamo (changes requested):
Please rebase your branch, we no longer have the `@animavita/models` package.

apps/backend/test/factories/adoption.ts
Vedu1996: Thank you for your feedback, I have addressed it now (hopefully)

belgamo (changes requested):
Please fix the broken tests. You can check if tests are breaking by running `pnpm backend test` and `pnpm backend test:e2e`

Vedu1996:
@belgamo Sorry :(, overlooked the actual values of the mock objects

belgamo (dismissed):
Thanks for the awesome PR!

Issue #200, opened 2023-09-26
Update sign up endpoint to return user's credentials


belgamo:
## Task
Currently, when signing up, we return a successful string message. But it's not useful, so let's return the user's credentials so we can auto-sign the user in on the front-end side.

## Requirements
- Returns the user's tokens
- Update tests

sivasuriyankumarasamy:
Hi @belgamo,
I would like to work on this.
Please assign this to me.

sivasuriyankumarasamy:
Hi @belgamo ,

I am getting below error. 
```
JwtStrategy requires a secret or key
TypeError: JwtStrategy requires a secret or key
```

How can I get ? 
```
JWT_ACCESS_SECRET=
JWT_REFRESH_SECRET=
```

belgamo:
> Hi @belgamo ,
> 
> I am getting below error. do I need to add any key?
> 
> `JwtStrategy requires a secret or key TypeError: JwtStrategy requires a secret or key`

Are you trying to run the app without docker? If so, please follow the [contribution guide](https://github.com/animavita/animavita/blob/v2/CONTRIBUTING.md#running-without-docker).

Basically, you have to set a secret in the `.env` file. If you're not familiar with that, you can also check [nest docs](https://docs.nestjs.com/security/authentication#jwt-token).

sivasuriyankumarasamy:
Thanks @belgamo.

I figured it out.

sivasuriyankumarasamy:
Hi @belgamo,

raised PR: https://github.com/animavita/animavita/pull/206

PR #202, opened 2023-10-02
Feat/add signup screen

### LINKED ISSUE #188: Add Signup screen
depends on #200

## Task
We already have the `GetLocation` screen. Let's continue the work started on https://github.com/animavita/animavita/pull/181 by creating the `SignUp` screen and integrating these two pieces to complete the user onboarding.

Endpoint -> `POST /api/v1/auth/signUp`

## Requirements
- Add the SignUp screen
- Persist user credentials on AuthContext 
- After signing up, the user lands on the app home screen
- Handle network errors
- Add tests

Isaac-alencar:
closes #188

## Sign up screen
This PR adds the sign up screen as part of the on boarding flow.

## Tests
I know it's missing some specs :( , so I need some guidance to write tests for this feature


apps/mobile/src/i18n/locales/pt-BR.json
```diff
@@ -61,6 +61,18 @@
...
     }
   },
+  "SIGN_UP": {
+    "FORM_ERROR_MESSAGES": {
+      "NAME_REQUIRED": "O seu nome é obrigatório",
+      "NAME_EMPTY": "O seu nome é obrigatório pode ser vazio",
+      "EMAIL_EMPTY": "Email não pode ser vazio",
+      "EMAIL_REQUIRED": "Email não pode ser vazio",
+      "EMAIL_EMAIL": "Digite um email válido",
+      "PASSWORD_EMPTY": "A senha não pode ser vazia",
```
belgamo, line 79: I think we can have only one message per field "A senha é obrigatória"

apps/mobile/src/providers/onboarding-provider/onboarding-provider.tsx
```diff
@@ -0,0 +1,63 @@
...
+  name: '',
+  password: '',
+  email: '',
+  location: {
+    latitude: 0,
+    longitude: 0,
+  },
+};
+
+export const OnBoardingContext = createContext<OnBoardingContextType>({} as OnBoardingContextType);
```
belgamo: I believe we could streamline this. Instead of using a context, WDYT about passing parameters through routes? https://reactnavigation.org/docs/params

apps/mobile/src/screens/signin/compose/signin-form.tsx
```diff
@@ -30,6 +34,14 @@ export const SignInForm = () => {
 
         <Input placeholder={t('SIGN_IN.FORM.PASSWORD_INPUT')} type="password" />
       </FormRow>
+      <Button
+        variant="link"
+        onPress={() => {
+          navigate(Routes.SignUp as never);
+        }}
+      >
+        Ou Registre-se
```
belgamo: Please move this to the translations

apps/mobile/src/screens/signup/compose/signup-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { FormField } from './form-field';
+import { UserType } from '../../../../../../shared/types';
+import { useSignUp } from '../../../hooks/use-onboarding-provider';
+import Routes from '../../../routes';
+
+type RegisterUserFormProps = {
+  defaultValues?: Partial<UserType>;
+};
+
+const createUserSchema = Joi.object({
```
belgamo: Please consume the existing schemas from `@animavita/validation-schemas`

apps/mobile/src/screens/signup/compose/signup-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <FormField label="Email" name="email" placeholder="Email" />
+          <FormField type="password" label="Senha" name="password" placeholder="Senha" />
+          <Button
+            marginTop={6}
+            width="full"
+            onPress={() => {
+              onConfirm(saveUserForm.getValues() as UserType);
+            }}
+          >
+            Registrar-se
```
belgamo: Please move this to the translations

apps/mobile/src/screens/signup/compose/signup-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <KeyboardAvoidingView
+      behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
+      style={{ flex: 1, marginTop: 50 }}
+      keyboardVerticalOffset={0}
+    >
+      <FormProvider {...saveUserForm}>
+        <FormControl isInvalid={!!saveUserForm.formState.errors}>
+          <FormField label="Nome completo" name="name" placeholder="Nome completo" />
```
belgamo: Please move this label to the translations

apps/mobile/src/screens/signup/compose/signup-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <FormField label="Email" name="email" placeholder="Email" />
+          <FormField type="password" label="Senha" name="password" placeholder="Senha" />
```
belgamo: Same

apps/mobile/src/screens/signup/signup.screen.tsx
```diff
@@ -0,0 +1,21 @@
...
+import theme from '../../theme';
+
+const SignUp = () => {
+  return (
+    <SafeArea>
+      <Stack flex="1" padding={6}>
+        <Heading fontSize="4xl" color={theme.colors.primary[600]}>
+          Animavita
+        </Heading>
+        <Text fontSize="2xl">Registrar-se</Text>
```
belgamo: Please move this to the translations

apps/mobile/src/screens/signup/compose/signup-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+      toast.show({
+        description: `Invalid data`,
+      });
+
+      return;
+    }
+
+    updateUserInfo(data);
+
+    navigate(Routes.GetLocation as never);
```
belgamo: So as mentioned earlier, you could get rid of the onboarding context and pass the user data filled up in the form through the routes parameters here.

belgamo (changes requested):

belgamo:
Don't forget to hit this endpoint `POST /api/v1/auth/signUp` to register the user after confirming the their location

apps/mobile/src/hooks/use-user-location/use-user-location.ts
```diff
@@ -1,6 +1,8 @@
 import * as Location from 'expo-location';
 import { useState } from 'react';
 
+import { Coordinates } from '../../../../../shared/types';
```
belgamo:
```suggestion
import { Coordinates } from '@animavita/shared/types';
```

apps/mobile/src/hooks/use-user-location/use-user-location.ts
```diff
@@ -13,6 +15,7 @@ export enum Warnings {
 
 const useUserLocation = () => {
   const [address, setAddress] = useState<Address>();
+  const [coords, setCoors] = useState<Coordinates>();
```
belgamo:
```suggestion
  const [coords, setCoords] = useState<Coordinates>();
```

apps/mobile/src/screens/get-location/get-location.screen.tsx
```diff
@@ -25,12 +24,17 @@ const errorAlert = (msg: string, onPress: () => void, text?: string) =>
   ]);
 
 const GetLocation = () => {
-  const auth = useAuth();
-  const { getLocation, address, isLoading, warning } = useGeolocation();
+  const { getLocation, address, isLoading, warning, coords } = useGeolocation();
   const { t } = useLocale();
-  const { navigate } = useNavigation();
+  const { params } = useRoute();
```
belgamo: `useRoute` accepts a generic for the `params` type.

apps/mobile/src/screens/get-location/get-location.screen.tsx
```diff
@@ -25,12 +24,17 @@ const errorAlert = (msg: string, onPress: () => void, text?: string) =>
...
 
-  const onConfirmLocation = () => navigate(Routes.Home as never);
+  const onConfirmLocation = () => {
+    const {
+      userInfo: { name, email, password },
+    } = params;
+
+    signUp({ name, email, password, coords });
```
belgamo: Please don't call the service function directly. Instead, instantiate [`useMutation`](https://tanstack.com/query/v4/docs/react/guides/mutations) and pass the service to it.

belgamo: Since we haven't finished https://github.com/animavita/animavita/issues/200 yet, please redirect the user to the login screen after the sign-up request succeeds.

apps/mobile/src/services/sign-up.ts
```diff
@@ -0,0 +1,7 @@
+import { UserType } from '@animavita/types';
+
+import client from './http-client';
+
+export const signUp = (user: UserType) => {
+  return client['post']('/auth/signUp', user);
```
belgamo:
```suggestion
  return client.post('/auth/signUp', user);
```

apps/mobile/src/services/sign-up.ts
```diff
@@ -0,0 +1,7 @@
+import { UserType } from '@animavita/types';
+
+import client from './http-client';
+
+export const signUp = (user: UserType) => {
```
belgamo: `UserType` doesn't seem to be the correct one to be used here. WDYT about creating a new `sign-up-dto.ts` file and a new `SignUpRequest` type in it? Similar to what we have in the `sign-in-dto` file.

apps/mobile/src/screens/signup/signup.screen.tsx
```diff
@@ -0,0 +1,21 @@
+import { Heading, Stack, Text } from 'native-base';
+
+import { SignUpForm } from './compose';
+import SafeArea from '../../components/safe-area';
+import theme from '../../theme';
+
+const SignUp = () => {
```
belgamo, line 7: I found a couple of problems on this screen:

1. There is no way to go back to the sign-in screen. Please add the left arrow to the top of the screen, similar to what we have on the profile  screen.
2. On iOS, the keyboard overlaps the form. Try to use the existing `SafeArea` component.
Simulator Screenshot - iPhone 14 Pro Max - 2023-10-12 at 12 28 10

apps/mobile/src/screens/signin/compose/signin-form.tsx
```diff
@@ -1,14 +1,18 @@
+import { useNavigation } from '@react-navigation/native';
 import { Button, Heading, Input, KeyboardAvoidingView, Text, View } from 'native-base';
 
 import { FormRow } from './form-row';
 import { useAuth } from '../../../hooks/use-auth-provider';
 import useLocale from '../../../hooks/use-locale';
+import Routes from '../../../routes';
 import theme from '../../../theme';
 
 export const SignInForm = () => {
```
belgamo, line 12: 1. When this component renders, the password field is the first one that gets focused. Please, make sure the first field (name) is the focused one.
2. When I focus on the field, it shows me the error state (red outline) right away. Check the register adoptions form to see how the fields were implemented in terms of error handling.

belgamo (changes requested):

apps/mobile/src/services/sign-up.ts
Isaac-alencar: Sure! I was gettin' type mismatch error and was about to ask to do the same. Thank you 😄

apps/mobile/src/screens/signup/signup.screen.tsx
belgamo: @Isaac-alencar instead of trying to solve this, what about making a multistep form? Like we have in the adoption form. On mobile apps, it's common to split big forms into separate steps. I think it's more user-friendly.

belgamo (changes requested):
1. Let's make a multistep form. Just like we have in the adoption form. For testing, we have a good example [here](https://github.com/animavita/animavita/blob/926a854ff83afa1728d5835e045da5f6d86f988c/apps/mobile/src/components/register-adoption-form/adoption-form.component.test.tsx).

2. I still see input style bugs when validation fails. If you want to use a label to indicate the error, I suggest you take a look here https://docs.nativebase.io/next/form-control#page-title. You can also use simple toasts messages, just like we have in the adoption form. 

3. https://github.com/animavita/animavita/pull/206 has been merged. You can now log the user in smoothly.

Let me know if you need any help :)

apps/mobile/src/screens/signup/signup.screen.tsx
Isaac-alencar: Thank you for reviewing! sounds good to me, I'll peek at the adoption form and implement the same approach

belgamo:
Since this PR has been open for too long, I don't want to spend more time on it. I pushed new commits with what I consider to be the most important fixes, including the one to make the form visible even when the keyboard is opened. I'm addressing a new issue https://github.com/animavita/animavita/issues/214 to finish the work started here.

PR #203, opened 2023-10-02
Add authorization with CASL

### LINKED ISSUE #159: Setup authorization
depends on https://github.com/animavita/animavita/pull/135

belgamo:
closes #159

## Overview
I started a POC on https://github.com/animavita/animavita/pull/184 but found [nest-casl](https://github.com/getjerry/nest-casl), which is an excellent library that has solved some of the issues I've struggled with.

## Changes
- Protect update adoption endpoint

### DIFF
--- a/apps/backend/src/adoption/adoption.controller.ts
+++ b/apps/backend/src/adoption/adoption.controller.ts
@@ -14,12 +14,15 @@ import {
   UseGuards,
   UsePipes,
 } from '@nestjs/common';
+import { AccessGuard, Actions, UseAbility } from 'nest-casl';
+
 import { JoiValidationPipe } from '../pipes/joi-validation-pipe';
 import { AdoptionsService } from './adoption.service';
 import { User } from '../decorators/user.decorator';
 import { AccessTokenGuard } from '../guards/accessToken.guard';
 import { JwtPayload } from '../auth/strategies/accessToken.strategy';
 import { CreateAdoptionRequest, UpdateAdoptionRequest } from '@animavita/types';
+import { AdoptionHook } from '../frameworks/casl/hooks/adoption.hook';
 
 @Controller('api/v1/adoptions')
 export class AdoptionsController {
@@ -36,7 +39,9 @@ export class AdoptionsController {
   }
 
   @Patch()
+  @UseGuards(AccessTokenGuard, AccessGuard)
   @UsePipes(new JoiValidationPipe(adoptionValidationSchema))
+  @UseAbility(Actions.update, 'Adoption' as any, AdoptionHook)
   async updateAdoption(@Body() adoption: UpdateAdoptionRequest) {
     return this.adoptionsService.updateAdoption(adoption);
   }
--- a/apps/backend/src/adoption/adoption.controller.ts
+++ b/apps/backend/src/adoption/adoption.controller.ts
@@ -23,6 +23,7 @@ import { AccessTokenGuard } from '../guards/accessToken.guard';
 import { JwtPayload } from '../auth/strategies/accessToken.strategy';
 import { CreateAdoptionRequest, UpdateAdoptionRequest } from '@animavita/types';
 import { AdoptionHook } from '../frameworks/casl/hooks/adoption.hook';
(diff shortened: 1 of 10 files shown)


apps/backend/src/adoption/adoption.controller.ts
```diff
@@ -36,7 +39,9 @@ export class AdoptionsController {
   }
 
   @Patch()
+  @UseGuards(AccessTokenGuard, AccessGuard)
   @UsePipes(new JoiValidationPipe(adoptionValidationSchema))
+  @UseAbility(Actions.update, 'Adoption' as any, AdoptionHook)
```
belgamo: Had to cast this on purpose probably due to this issue https://github.com/getjerry/nest-casl/issues/349

apps/backend/src/frameworks/casl/app.roles.ts
```diff
@@ -0,0 +1,5 @@
+export enum Roles {
+  Admin = 'admin',
+  Owner = 'owner',
+  User = 'user',
+}
```
petruspierre, line 5: Why is having this distinction between User/Owner relevant?

apps/backend/src/frameworks/casl/hooks/user.hook.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable } from '@nestjs/common';
+import { UserBeforeFilterHook } from 'nest-casl';
+
+import { DataServices } from '../../../core/abstracts/data-services.abstract';
+
+@Injectable()
+export class UserHook implements UserBeforeFilterHook<any> {
+  constructor(readonly dataServices: DataServices) {}
+
+  async run(user: any) {
```
petruspierre: Do we need this `any` here?

apps/backend/src/frameworks/casl/hooks/user.hook.ts
```diff
@@ -0,0 +1,26 @@
...
+    const dbUser = await this.dataServices.users.findByEmail(user.email);
+
+    // TO DO:
+    // since nest-casl is RABC, we need to generate an array of roles.
```
petruspierre: RBAC*

Also, what about filing an issue and link it to the comment so we can remember removing it in the future?

apps/backend/src/adoption/adoption.controller.ts
petruspierre: I see we use 'Adoption' in another place of the code, what do you think about having that separately on an enum or something?

petruspierre (commented):
nest-casl is nice! I left you some comments

apps/backend/src/adoption/adoption.controller.ts
belgamo: Yep, nice catch! I'll centralize and reuse it. Thanks

apps/backend/src/frameworks/casl/app.roles.ts
belgamo: Basically, an owner can't adopt pets. By default, every new joiner would be considered a user. Once they post adoptions, they become owners. IMO, a user who is posting pets for adoption can't adopt other pets. Otherwise, we might leave an open door for pet exchange, which is not what we want.

apps/backend/src/frameworks/casl/hooks/user.hook.ts
belgamo: done! https://github.com/animavita/animavita/issues/204

apps/backend/src/frameworks/casl/hooks/user.hook.ts
belgamo: Nope

belgamo:
We still cannot perform restricted field validation effortlessly. I've opened a PR on nest-casl https://github.com/getjerry/nest-casl/pull/761

apps/backend/src/frameworks/casl/app.roles.ts
brunohkbx: @belgamo I'd call that out in this module for future reference. I think adding some documentation at the top would be helpful for onboarding new folks in this project

PR #206, opened 2023-10-12
Updates signup endpoint to return refresh token

sivasuriyankumarasamy:
Hi @belgamo, 

Raised PR for  #200 

Kindly review and let me know If anything.


belgamo (approved):
Awesome, thank you so much!

Issue #158, opened 2023-02-10
Add swagger/docs for backend API


belgamo:

jrsmarcilio:
Hello @belgamo, I would like to work on this issue. Can you assign it to me?

PR #211, opened 2023-10-27
Add swagger/docs for backend API

jrsmarcilio:
#158


apps/backend/src/swagger/swagger.config.ts
belgamo, line 1: Could you place this file under `frameworks/swagger`?

apps/backend/src/main.ts
```diff
@@ -1,9 +1,32 @@
 import { NestFactory } from '@nestjs/core';
+import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
 import { AppModule } from './app.module';
+import { swaggerDocumentOptions } from './swagger/swagger.config';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
+  const swaggerSettings = new DocumentBuilder()
```
belgamo: In the file under the `frameworks` folder, create a class to abstract the whole necessary logic to start Swagger. Then you call it here

apps/backend/src/main.ts
```diff
@@ -1,9 +1,32 @@
...
+    .setTitle('Swagger Animavita')
+    .setDescription(
+      'This is a sample server Animavita server. You can find out more about Swagger at [http://swagger.io](http://swagger.io).',
+    )
+    .setVersion('1.0.0')
```
belgamo: I think here you're supposed to use `SWAGGER_VERSION` constant

apps/backend/package.json
belgamo, line 1: Looks like you forgot to commit the lock file

belgamo (changes requested):
I haven't had a chance to go through all the files yet. Therefore, I need to play around to make sure the DTOs you created are necessary and will fit into our current architecture.

apps/backend/src/swagger/swagger.config.ts
jrsmarcilio: Done!

apps/backend/src/main.ts
jrsmarcilio: Done!

apps/backend/src/main.ts
jrsmarcilio: Done!

apps/backend/package.json
jrsmarcilio: Done!

apps/backend/src/adoption/adoption.controller.ts
```diff
@@ -14,44 +14,81 @@ import {
...
 
+  @ApiTags('Adoptions')
+  @ApiOperation({ summary: 'Search near me' })
+  @ApiOkResponse({ description: 'Search successfully' })
+  @ApiNotFoundResponse({ description: 'Adoption not found' })
+  @ApiUnprocessableEntityResponse({ description: 'Bad Request' })
+  @ApiForbiddenResponse({ description: 'Unauthorized Request' })
   @UseGuards(AccessTokenGuard)
   @Get('nearMe')
   async findNearMe(
```
belgamo, line 94: There's no way to provide the radius range to find near adoptions 
Screenshot 2023-12-03 at 14 03 06

belgamo:
note for me: update CI env config with the `SWAGGER_VERSION` variable.

belgamo:
closing due to inactivity

PR #227, opened 2024-03-27
Fix/user location web issue

RafaelAllves:
Proposed solution: Use [Opencage](https://opencagedata.com/) for reverse geocoding service on the web version.


CONTRIBUTING.md
```diff
@@ -78,6 +78,8 @@ To point the app to staging, create a .env file and set the `ENV` variable to `s
 cp .env.example .env
 ```
 
+To use the geolocation service when running on the web version, you need to have an API key from Opencage. You can obtain an API key for free by registering on their website, and there is no need to register a payment method for this. After obtaining the API key, add it to the `OPENCAGE_KEY` environment variable.
```
belgamo: could you move this above? close to the backend section

apps/backend/src/opencage/opencage.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+      https.get(url, (response) => {
+        let data = '';
+
+        response.on('data', (chunk) => {
+          data += chunk;
+        });
+
+        response.on('end', () => {
+
+          resolve(JSON.parse(data)?.results.map(address => ({region: address.components.city, subregion: address.components.state})));
```
belgamo: i live in Portugal and i ended up getting this Screenshot 2024-03-29 at 11 18 19

this works perfectly with Brazilian addresses and our app is first supposed to work only in Brazil. however, we do have plans to expand it abroad so it would be great to support other formats. I'm not sure about all formats, but I believe for now we could fall back the city to `municipality` and the state to `county`


Screenshot 2024-03-29 at 11 24 31

apps/backend/src/opencage/opencage.controller.ts
```diff
@@ -0,0 +1,31 @@
+import { Controller, Get, Query } from '@nestjs/common';
+import * as https from 'https';
+
+@Controller('api/v1/location')
+export class OpencageController {
```
belgamo: i'd rather hide from our code which service we're using. i also think this controller is holding too much logic that it's not responsible for. WDYT about creating a `GeolocationService` interface and making a concrete implementation for opencage? then our code would only know this interface, not the implementation details. you can use dependency injection to provide the implementation.

it's gonna help us to keep the core uncoupled from details if at some point we wish to replace opencage, we just need to create a new implementation and replace it in the DI container without having to touch on the main code.

apps/backend/src/opencage/opencage.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Get()
+  async reverseGeocoding(@Query('latitude') latitude: string, @Query('longitude') longitude: string): Promise<any> {
+    if (!latitude || !longitude) {
+      throw new Error('Latitude and longitude must be provided.');
+    }
+
+    const url = `https://api.opencagedata.com/geocode/v1/json?q=${latitude}+${longitude}&key=${process.env.OPENCAGE_KEY}`;
+
+    return new Promise((resolve, reject) => {
+      https.get(url, (response) => {
```
belgamo: nest has a built-in http module you could use instead https://docs.nestjs.com/techniques/http-module#http-module

apps/backend/src/opencage/opencage.controller.ts
belgamo: also, the opencage implementation should live under the `frameworks` folder

belgamo (changes requested):
Thank you so much for that, you nailed it! I just pointed out a few changes before merging.

CONTRIBUTING.md
RafaelAllves: It was unclear whether this content should be in the "Running the infrastructure & backend with Docker" section or the "Running without Docker" section.

apps/backend/src/opencage/opencage.controller.ts
RafaelAllves: From what I understand, there is a difference in the way cities and states are referenced here in South America compared to Europe. Here, the terms "municipality" and "county" generally refer to metropolitan regions, which could cause confusion. Therefore, I chose a logical approach that works correctly on both continents, taking this distinction into account.

apps/backend/src/opencage/opencage.controller.ts
RafaelAllves: I tried to create a GeolocationService interface and use the following snippet in the GeolocationModule providers:


``` typescript
providers: [
    { provide: GeolocationService, useClass: OpenCageService },
],
```
However, this did not work and resulted in the following error:

```
'GeolocationService' only refers to a type, but is being used as a value here.
```
Although I was unable to resolve this error, I was still able to refactor the code to improve it.

apps/backend/src/opencage/opencage.controller.ts
RafaelAllves: I am experiencing the following error when trying to use the `@nestjs/axios` package: "Cannot find module `@nestjs/axios' or its corresponding type declarations."

I tried several approaches to install and use `@nestjs/axios`, but I was unable to resolve the issue. I believe that my lack of experience with PNPM may be contributing to this.

If you could offer any guidance or suggestions on how to resolve this issue, I would greatly appreciate the help.

apps/backend/src/opencage/opencage.controller.ts
belgamo: Yeah but it looks like there's no city and state properties available in the return payload for European addresses

apps/backend/src/opencage/opencage.controller.ts
belgamo: I'll play around with it and let you know

apps/backend/src/opencage/opencage.controller.ts
belgamo: Sure, I'll take a look and maybe we can hop on a call on discord and pair on it

belgamo: so looks like we have a problem on our docker configuration. every time we add a new dependency it's expected we have to rebuild the container because the `node_modules` is cached, so something like `docker-compose up --rebuild` should work. however, I only got this working when I manually deleted the container and ran the command again. I'll take a look at it later, but for now you can:

1. install the dependencies with `pnpm add -D @nestjs/axios axios --filter=backend`
2. delete the container manually
3. `docker-compose up`

apps/backend/src/opencage/opencage.controller.ts
belgamo: so you probably tried to define`GeolocationService` as an interface right? by interface I meant something like an abstract class because by creating a class you can also use it as the key in the DI container. there's a `core/abstracts` folder where you can create something like this:

```js
interface ReversedGeocodingOutput {
  region: string;
  subregion: string;
}

export abstract class GeolocationService {
  public abstract reverseGeocoding(
    latitude: string,
    longitude: string,
  ): Promise<ReversedGeocodingOutput>;
}
```

then just extend it in the implementation:

```
@Injectable()
export class OpenCageService extends GeolocationService { }
```

it's important to mark the implementation as injectable so the container can inject the HTTP service later for you.

CONTRIBUTING.md
```diff
@@ -53,6 +53,8 @@ Alternatively you can setup your own mongo database without docker. Then, from t
 $ cp apps/backend/.env.example apps/backend/.env
 ```
 
+To use the geolocation service when running on the web version, you need to have an API key from Opencage. You can obtain an API key for free by registering on their website, and there is no need to register a payment method for this. After obtaining the API key, add it to the `OPENCAGE_KEY` environment variable.
```
belgamo:
```suggestion
To use the geolocation service when running the mobile app on the web version, you need to have an API key from Opencage. You can obtain an API key for free by registering on their website, and there is no need to register a payment method for this. After obtaining the API key, add it to the `OPENCAGE_KEY` environment variable.
```

apps/backend/src/frameworks/opencage.service.ts
```diff
@@ -0,0 +1,44 @@
...
+                return {
+                  region: address.components.municipality,
+                  subregion: address.components.county,
+                }
+              } else {
+                return {
+                  region: address.components.city,
+                  subregion: address.components.state,
+                }
+              }
```
belgamo: hmmm i don't know... maybe we could just use the `municipality` and `county` as simple fallbacks for everything, not just Europe? WDYT?

```js
const { 
 city,
 state, 
 municipality,
 county
} = address.components;

return {
 region: city || municipality,
 subregion: state || county
};

File: apps/backend/src/frameworks/opencage.service.ts [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+
+export class OpenCageService {
+  async reverseGeocoding(latitude: string, longitude: string): Promise<any> {
+    if (!latitude || !longitude) {
+      throw new Error('Latitude and longitude must be provided.');
+    }
+
+    const url = `https://api.opencagedata.com/geocode/v1/json?q=${latitude}+${longitude}&key=${process.env.OPENCAGE_KEY}`;
+
+    return new Promise((resolve, reject) => {
```
belgamo: once you replace that with the built-in HTTP module, please use async/await instead of promises.

apps/backend/src/geolocation/geolocation.controller.ts
```diff
@@ -0,0 +1,12 @@
+import { Controller, Get, Query } from '@nestjs/common';
+import { OpenCageService } from '../frameworks/opencage.service';
+
+@Controller('api/v1/location')
+export class GeolocationController {
+  constructor(private readonly geolocationService: OpenCageService) {}
```
belgamo: the controller shouldn't know about which service we're using. instead, we should only depend on the abstract class you created.

apps/backend/src/frameworks/opencage.service.ts
```diff
@@ -0,0 +1,33 @@
+import { Injectable } from '@nestjs/common';
+import { HttpService } from '@nestjs/axios';
+import { GeolocationService } from 'src/geolocation/geolocation.service';
```
belgamo: i think it might break our test.

```suggestion
import { GeolocationService } from '../core/abstracts/geolocation-service.abstract';
```

apps/backend/src/geolocation/geolocation.module.ts
```diff
@@ -0,0 +1,12 @@
+import { Module } from '@nestjs/common';
+import { GeolocationController } from './geolocation.controller';
+import { OpenCageService } from '../frameworks/opencage.service';
+import { GeolocationService } from './geolocation.service';
+import { HttpModule } from '@nestjs/axios';
+
+@Module({
+  imports: [HttpModule],
+  controllers: [GeolocationController],
+  providers: [OpenCageService],
```
belgamo:
```suggestion
  providers: [
    {
      provide: GeolocationService,
      useClass: OpenCageService,
    },
  ]
```

apps/backend/src/geolocation/geolocation.service.ts
belgamo, line 1: please move this to `core/abstracts`

belgamo (changes requested):

belgamo (approved):

belgamo:
Good job @RafaelAllves!

PR #228, opened 2024-03-30
update docker instructions

belgamo:
solves https://github.com/animavita/animavita/pull/227#discussion_r1544696321

References:

https://jaygould.co.uk/2022-08-12-docker-node-problems-development-prisma/

https://github.com/docker/compose/issues/4337

https://docs.docker.com/storage/bind-mounts/#use-a-bind-mount-with-compose

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,11 @@ We have a `docker-compose` file that sets up a mongodb database and the backend 
 $ docker-compose up -d
 ```
 
-> Keep in mind if you change the shared packages you must rebuild the docker image.
+We're caching the `node_modules` folder to leverage cross-platform compatibility. So the first time the container is up, an anonymous volume gets created for the dependencies. With that said, whenever you change the shared packages or install a new dependency, you must rebuild the docker image telling docker to **NOT** use the existing volume from the previous container:
+
+```sh
+$ docker-compose up --build --force-recreate -V
+```
 
 ### Running without docker
 


belgamo:
@RafaelAllves could you review and check if it solves your problem when adding a new dependency?

RafaelAllves:
After thoroughly reviewing and testing, I confirm that this approach effectively resolves the issue. However, during testing I noticed some interesting behavior:

After installing a new dependency, I run the command:

```sh
docker-compose up --build
```
As expected, this resulted in the previously mentioned error. Then I run:

```sh
docker-compose up --build --force-recreate -V
```
And this resolved the issue as expected. However, I decided to perform deeper testing and completely removed the previously installed dependency as well as manually deleting the Docker containers.

After reinstalling the dependency and trying to run again:

```sh
docker-compose up --build
```
I expected the error to occur again. However, to my surprise, the installation completed correctly, without the need to use --force-recreate -V.

Although I recorded this curious behavior, I reinforce that the initial approach effectively resolved the problem.

RafaelAllves (approved):

belgamo:
> After thoroughly reviewing and testing, I confirm that this approach effectively resolves the issue. However, during testing I noticed some interesting behavior:
> 
> After installing a new dependency, I run the command:
> 
> ```shell
> docker-compose up --build
> ```
> 
> As expected, this resulted in the previously mentioned error. Then I run:
> 
> ```shell
> docker-compose up --build --force-recreate -V
> ```
> 
> And this resolved the issue as expected. However, I decided to perform deeper testing and completely removed the previously installed dependency as well as manually deleting the Docker containers.
> 
> After reinstalling the dependency and trying to run again:
> 
> ```shell
> docker-compose up --build
> ```
> 
> I expected the error to occur again. However, to my surprise, the installation completed correctly, without the need to use --force-recreate -V.
> 
> Although I recorded this curious behavior, I reinforce that the initial approach effectively resolved the problem.

it's because docker caches the previously builded images. so if you try to build an image that has been already builded before, it's expected to work. same happened here. if we try to do the same with a flag to avoid caching or something or try installing a different dependency, the issue should occur anyway

PR #229, opened 2024-04-17
Fix nodejs version in backend/Dockerfile

gustavonobreza:
fix(docker): nodejs version update 16v to 18v.

pnpm@9v in the latest version require node@18v as necessary dependency. https://pnpm.io/installation#compatibility

### DIFF
--- a/apps/backend/Dockerfile
+++ b/apps/backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16 AS base
+FROM node:18 AS base
 RUN npm i -g pnpm
 
 FROM base AS dependencies


belgamo (changes requested):
Thanks for your contribution!

Honestly, I think it's time to bump up the node version to the LTS one, I think both nest and expo support v20. Would you mind working on that? Please remember to update the `CONTRIBUTING.md` and `nvmrc` files as well as running `pnpm install` so the lock file gets updated. Thanks.

belgamo:
closing due to inactivity

PR #243, opened 2025-01-05
Backend architecture refactoring: post pet for adoption

belgamo:
In v2, we've been trying to follow Clean Architecture & DDD. Some misconceptions were happening, resulting in a very complex and over-engineered code. I'm starting to clean this up and get rid of what doesn't make sense.

This PR adds an actual domain layer (core) as well as an application layer (use cases) and proper error handling by mapping domain errors to HTTP responses. Missing concepts were added such as pet, size, type, etc, and are represented by domain objects and value objects. It also improves our test suite by implementing actual unit and integration tests.


apps/backend/src/domain/location/location.ts
```diff
@@ -0,0 +1,24 @@
+import { EntityError } from '../errors/entity.error';
+
+export default class Location {
+  private value: {
+    latitude: number;
+    longitude: number;
+  };
+
+  constructor(longitude: number, latitude: number) {
+    if (typeof latitude !== 'number') {
```
rrambul, line 10: Latitude and longitude have natural constraints (latitude: -90 to 90, longitude: -180 to 180). Adding range checks can prevent invalid coordinates.

apps/backend/src/usecases/owner/get-pet/get-pet.ts
```diff
@@ -0,0 +1,38 @@
...
+  size: string;
+  observations: string;
+  photos: string[];
+};
+
+@Injectable()
+export default class GetPet {
+  constructor(
+    @InjectModel(MongoPet.name)
+    private readonly petModel: Model<PetDocument>,
```
rrambul, line 21: Instead of passing the Model directly to the useCase, you could create a repository that uses Mongoose to access the data and pass that repository to the useCase, ensuring that the application's useCase isn't dependent on Mongoose. This approach also allows you to swap in another repository that uses a different ORM to access the data in the future if necessary. Does that make sense?

apps/backend/src/usecases/owner/post-pet-for-adoption/post-pet-for-adoption.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { INestApplication } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import {
+  closeInMongodConnection,
+  TestMongoDataServicesModule,
+} from '../../../../test/utils/in-memory-mongo';
```
rrambul, line 6: What do you think about using path aliases and barrel files to simplify and better organize imports?

RenanRSilva (commented):
I have some questions and suggestions

apps/backend/src/usecases/owner/get-pet/get-pet.ts
belgamo: thanks for commenting! the "model" here is a database model, not an entity model. the repository pattern doesn't play well here because this use case is a reading model, it just grabs data and returns it to whoever needs it. the repository pattern is directly related to how aggregates are persisted, so here we're not mutating nor persisting anything.

with regards to uncoupling it from mongoose, I totally agree and have done so for the `find-nearest-pets`. since `get-pet` is just being used for testing, I skipped this for now.

apps/backend/src/usecases/owner/post-pet-for-adoption/post-pet-for-adoption.spec.ts
belgamo: yep, that's a good idea! just haven't had the time yet, but feel free to contribute :D

PR #244, opened 2025-01-12
Backend architecture refactoring: find nearest pets usecase

belgamo:
This PR refactors the work done in https://github.com/animavita/animavita/pull/173

Had to rollback the node version used by the action that runs backend tests due to https://github.com/typegoose/mongodb-memory-server/issues/732

### DIFF
--- a/apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
+++ b/apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
@@ -9,7 +9,7 @@ type Input = {
 };
 
 @Injectable()
-export default class FindNearestPet {
+export default class FindNearestPets {
   constructor(
     @Inject(PET_DAO) private readonly petDAO: PetDao,
     @Inject(UserService) private readonly userService: UserRepository,
--- a/apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
+++ b/apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
@@ -1,0 +1,29 @@
+import { Inject, Injectable } from '@nestjs/common';
+import { UserService } from '../../../../user/user.service';
+import { UserRepository } from '../../../../user/repositories/user-repository.interface';
+import { PET_DAO, PetDao } from '../../../dao/pet.dao';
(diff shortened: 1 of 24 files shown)


apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
```diff
@@ -0,0 +1,29 @@
+import { Inject, Injectable } from '@nestjs/common';
+import { UserService } from '../../../../user/user.service';
+import { UserRepository } from '../../../../user/repositories/user-repository.interface';
```
rrambul, line 3: What do you think about using path aliases and barrel files to simplify and better organize imports?

apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
rrambul, line 1: Great job! You removed the service method, introduced the useCase, DAO, DTO, and added tests. I believe this is a significant improvement!

apps/backend/src/infra/dao/mongo-pet.dao.ts
rrambul, line 1: What is the responsibility of a DAO (Data Access Object) in the architecture?

belgamo: DAOs are a better fit for read-only operations. why? because they’re all about grabbing data straight from the database without dragging in all the extra stuff that comes with domain entities.

repositories, on the other hand, are built to deal with full-blown entities. they’re great when you need to update or save things while keeping all the business rules in check. but using repositories just to read data is overkill. you’d often need to juggle multiple repositories to get what you need, which adds unnecessary bureacraucy.

with a DAO, it’s much simpler. you just fetch the data you want, usually as plain objects or DTOs, and skip all the extra steps. it’s quick, clean, and perfect for straightforward read operations.

apps/backend/src/application/usecases/adopter/find-nearest-pets/find-nearest-pets.ts
belgamo: i think it improves readability and helps to better organize the imports. i'll leave it up to someone looking for a good first issue

Issue #204, opened 2023-10-03
Add user roles


belgamo:
## Task
There will be three basic roles: `admin`, `adopter`, and `owner` (you can come up with better terms if you want). Regular users can select what they are (adopter or owner) when signing up through a screen that will be added on the frontend. We will manually add admins users.

### Motivation for telling users apart
A user who is posting pets for adoption can't adopt other pets. Otherwise, we might leave an open door for pet exchange, which is not what we want. In addition, we want the adoptions to go through an admin approval step to avoid fake posts in the app.

On the frontend, there will be different experiences based on the user role. On the backend, we might use that to allow or disallow user actions.

- [x] Update owner use cases to throw if user is an adopter
- [ ] Update adopter use cases to throw if user is an owner
- [x] Add tests

leossantiago:
I'm interested on working on this issue. @belgamo could you assign it to me?

belgamo:
Hi @leossantiago, any updates here? Can I assign it to someone else?

PR #266, opened 2025-06-08
feat: add user roles

belgamo:
closes https://github.com/animavita/animavita/issues/204


RenanRSilva (approved):
Nice work! One suggestion: right now, the allowed roles ('adopter' and 'owner') are hardcoded in assignRole. It might be better to define these roles in a central config or an enum so they're easier to manage and update.

